PR #2, opened 2019-02-28
Issue/jperf 405 thread dumps

wyrzyk:


.travis.yml
```diff
@@ -8,7 +8,7 @@ stages:
 jobs:
   include:
     - stage: test
-      script: travis_wait 14 ./gradlew build
+      script: travis_wait 16 ./gradlew build
```
sebapawlak: Anything we can do about it? Maybe run tests in parallel?

wyrzyk: > Anything we can do about it? Maybe run tests in parallel?

I've tried to bump maxParallelForks but it didn't help. We can do the same trick we do for our internal tests and take full control over the test's concurrency.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/ThreadDump.kt
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Creates Thread Dumps on a remote server over an ssh connection.
+ *
+ * @param pid process identifier.
+ * @param jdk java development kit.
+ *
+ * @since 4.9.0
+ */
+class ThreadDump(
+    private val pid: Int,
```
dagguh: AFAIK `Profiler` SPI already fitted the use case

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/ThreadDump.kt
```diff
@@ -0,0 +1,33 @@
...
+     * in [destination] folder.
+     *
+     * @param connection ssh connection to a remote machine.
+     * @param destination destination folder for thread dumps.
+     *
+     * @since 4.9.0
+     */
+    fun gather(connection: SshConnection, destination: String) {
+        val threadDumpName = Instant.now().toEpochMilli()
+        val command = "${jdk.use()}; jcmd $pid Thread.print > $destination/${threadDumpName}"
```
dagguh: Atlassian recommends using `jstack -l $pid`: https://confluence.atlassian.com/adminjiraserver073/generating-a-thread-dump-861253873.html

Does this implementation include info about `Locked ownable synchronizers`?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/ThreadDump.kt
wyrzyk: Sorry, I don't understand. Profiler and Thread Dumps are two separate things and I don't get your comment. Can you elaborate?

I'd like to gather thread dumps only on Jira startup, as it already bites us multiple times. Gathering thread dumps during the test would affect test results.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/ThreadDump.kt
wyrzyk: > The release of JDK 8 introduced Java Mission Control, Java Flight Recorder, and jcmd utility for diagnosing problems with JVM and Java applications. It is suggested to use the latest utility, jcmd instead of the previous jstack utility for enhanced diagnostics and reduced performance overhead.

https://docs.oracle.com/javase/8/docs/technotes/guides/troubleshoot/tooldescr016.html

> Prints Java thread stack traces for a Java process, core file, or remote debug server. This command is experimental and unsupported.

https://docs.oracle.com/javase/8/docs/technotes/tools/unix/jstack.html

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/ThreadDump.kt
dagguh: > Profiler and Thread Dumps are two separate things and I don't get your comment. Can you elaborate?

Not really. `AsyncProfiler` provides the same kind of data, but on a more frequent matter with smaller overhead. If the user needs thread dumps from the run, then they have to [write boilerplate to inject thread dumps via the `Profiler` SPI](https://github.com/atlassian/jira-hardware-exploration/commit/cf709ec37ee46d0591ec7eb0c7e76269872baccc).

> Gathering thread dumps during the test would affect test results.

It would or it would not, it's a judgement call on the user side. The SPI is optional, we don't enforce thread dumps by default. Also, when you have poor performance and CPU are idle, then thread dumps are most consistent way of drilling into thread liveliness issues.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/ThreadDump.kt
dagguh: > The -l option, which instructs the utility to look for ownable synchronizers in the heap and print information about java.util.concurrent.locks. Without this option, the thread dump includes information only on monitors.

Does jcmd have this? How does an example thread dump look like?

wyrzyk: Yes, jcmd has the option, but it's off by default.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/ThreadDump.kt
wyrzyk: That's an excellent example of why we shouldn't provide ThreadDump as a Profiler. It would limit the usage only to profiling the runtime. Now with a little boilerplate, you can use it as a profiler or as a standalone tool. We can later add ThreadDumpProfiler that does the thread dumps in the interval. (It would be misleading to pass a duration parameter to ThreadDump object, and it'll be obvious why you're passing a period to the ThreadDumpProfiler).

dagguh: Ok, I think it's clear to me now. So the way I see it, `ThreadDump` is a standalone component, which might be useful for various `Profiler` implementations. We might or might not provide a `ThreadDumpProfiler`, but that's out of scope for this PR, because the use case is different.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/ThreadDump.kt
dagguh: Ok, so we can switch on the option if we need it in the future.

PR #3, opened 2019-03-05
JPERF-408 Set GC params based on Java version

masiorski:
To be able to run performance tests for OpenJDK 11 we have to set GC params based on Java version.

### DIFF
--- a/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/JiraJvmArgs.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/JiraJvmArgs.kt
@@ -13,17 +13,43 @@ class JiraJvmArgs(
         debug: JvmDebug,
         jmx: RemoteJmx,
         jiraIp: String
-    ): List<JvmArg> = listOf(
-        JvmArg("-Datlassian.darkfeature.jira.onboarding.feature.disabled=", "true"),
-        JvmArg("-Djira.startup.warnings.disable=", "true"),
-        JvmArg("-XX:+PrintGCDetails"),
-        JvmArg("-XX:+PrintGCDateStamps"),
-        JvmArg("-XX:+PrintGCTimeStamps"),
-        JvmArg("-XX:+PrintGCCause"),
-        JvmArg("-XX:+PrintTenuringDistribution"),
-        JvmArg("-XX:+PrintGCApplicationStoppedTime"),
-        JvmArg("-XX:+UseGCLogFileRotation"),
-        JvmArg("-XX:NumberOfGCLogFiles=", "5"),
-        JvmArg("-XX:GCLogFileSize=", "20M")
-    ) + debug.getJvmOption() + jmx.getJvmArgs(jiraIp) + extra
+    ): List<JvmArg> = java8Arguments() + debug.getJvmOption() + jmx.getJvmArgs(jiraIp) + extra
+
+    fun arguments(
+        debug: JvmDebug,
+        jmx: RemoteJmx,
+        jiraIp: String,
+        jdkVersion: Int
+    ): List<JvmArg> {
+        val jvmArg = if (jdkVersion > 8) java9Arguments() else java8Arguments()
+        return jvmArg + debug.getJvmOption() + jmx.getJvmArgs(jiraIp) + extra
+    }
+
+    private fun java8Arguments(): List<JvmArg> {
+        return commonArguments() + listOf(
+            JvmArg("-XX:+PrintGCDetails"),
+            JvmArg("-XX:+PrintGCDateStamps"),
+            JvmArg("-XX:+PrintGCTimeStamps"),
+            JvmArg("-XX:+PrintGCCause"),
+            JvmArg("-XX:+PrintTenuringDistribution"),
+            JvmArg("-XX:+PrintGCApplicationStoppedTime"),
+            JvmArg("-XX:+UseGCLogFileRotation"),
+            JvmArg("-XX:NumberOfGCLogFiles=", "5"),
+            JvmArg("-XX:GCLogFileSize=", "20M")
+        )
+    }
+
+    private fun commonArguments(): List<JvmArg> {
+        return listOf(
+            JvmArg("-Datlassian.darkfeature.jira.onboarding.feature.disabled=", "true"),
+            JvmArg("-Djira.startup.warnings.disable=", "true")
+        )
+    }
+
+    private fun java9Arguments(): List<JvmArg> {
+        return commonArguments() + listOf(
+            JvmArg("-Xlog:gc*:file=atlassian-jira-gc-%t.log:time,uptime:filecount=5,filesize=20M")
+        )
+    }
+
 }
--- a/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/SetenvSh.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/SetenvSh.kt
@@ -3,6 +3,7 @@ package com.atlassian.performance.tools.infrastructure.api.jira
 import com.atlassian.performance.tools.infrastructure.api.Sed
 import com.atlassian.performance.tools.infrastructure.api.jira.SetenvSh.Variables.*
 import com.atlassian.performance.tools.infrastructure.api.jvm.JvmArg
+import com.atlassian.performance.tools.infrastructure.api.jvm.VersionedJavaDevelopmentKit
 import com.atlassian.performance.tools.ssh.api.SshConnection
 
 /**
@@ -28,14 +29,17 @@ class SetenvSh(
         gcLog: JiraGcLog,
         jiraIp: String
     ) {
+        val jdkVersion = if (config.jdk is VersionedJavaDevelopmentKit) config.jdk.getMajorVersion() else 8
         val args = config.jvmArgs
         val original = connection.execute("cat ${this.location}").output
         setMemory(connection, args, original)
-        val jvmArgs = args.arguments(
+        var jvmArgs = args.arguments(
             debug = config.debug,
             jmx = config.remoteJmx,
-            jiraIp = jiraIp
-        ) + gcLog.jvmArg()
+            jiraIp = jiraIp,
+            jdkVersion = jdkVersion
+        )
+        if (jdkVersion == 8) jvmArgs += gcLog.jvmArg()
         setArguments(connection, jvmArgs, original)
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,10 @@ Adding a requirement of a major version of a dependency is breaking a contract.
 Dropping a requirement of a major version of a dependency is a new contract.
 
 ## [Unreleased]
-[Unreleased]: https://github.com/atlassian/infrastructure/compare/release-4.10.0...master
+[Unreleased]: https://github.com/atlassian/infrastructure/compare/release-4.11.0...master
+
+## [4.11.0] - 2019-03-12
+[4.11.0]: https://github.com/atlassian/infrastructure/compare/release-4.10.0...release-4.11.0
 
 ### Added
 - Set GC params based on Java version. Resolve [JPERF-408].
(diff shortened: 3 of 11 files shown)


src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/SetenvSh.kt
```diff
@@ -28,14 +28,19 @@ class SetenvSh(
         gcLog: JiraGcLog,
         jiraIp: String
     ) {
+        val jdk = config.jdk
+        var jdkVersion = 8
+        if (jdk::class.qualifiedName?.contains("11") ?: false) jdkVersion = 11
```
dagguh: ## How to detect JDK version?
So we want to learn which Java version is used on the remote Jira node. There are many ways to achieve this.

### Class name
This PR currently relies on an assumption that any `JavaDevelopmentKit` with Java 11 will have `11` in its class name. That's a very hidden and implicit assumption, which makes it brittle.

### Explicit SPI
We might require an explicit declaration of the JDK version. It's hard to break, but it requires new SPI.

### `java -version`
We could actually run `java -version` via SSH and parse the result. I'm not sure how consistent the output format is across JDKs, so we might have to add new "parse version" SPI.

## How to expand SPI?
`JavaDevelopmentKit` is a public SPI, so some implementations are outside of our control. For example, our internal teams already experimented with AdoptJDK and other JDKs.
Simply adding a method to an interface will make the external implementations fail to compile and fail at runtime.
Currently, the best way to do that is to actually port the interface to Java, as we did before: https://github.com/atlassian/jira-actions/commit/0be33de0c3fd852c244e1180c6281caf660baaaa

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/SetenvSh.kt
```diff
@@ -28,14 +28,19 @@ class SetenvSh(
...
-        val jvmArgs = args.arguments(
+        var jvmArgs = args.arguments(
             debug = config.debug,
             jmx = config.remoteJmx,
-            jiraIp = jiraIp
-        ) + gcLog.jvmArg()
+            jiraIp = jiraIp,
+            jdkVersion = jdkVersion
+        )
+        if (jdkVersion == 8) jvmArgs += gcLog.jvmArg()
```
dagguh: It seems to say that we only support GC logs for JDK 8. If so, why?
Are GC params different across versions? Across JDKs?
Perhaps the GC params are so JDK-specific that they should be a part of the `JavaDevelopmentKit` interface?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/JiraJvmArgs.kt
```diff
@@ -13,17 +13,44 @@ class JiraJvmArgs(
...
-        JvmArg("-XX:NumberOfGCLogFiles=", "5"),
-        JvmArg("-XX:GCLogFileSize=", "20M")
-    ) + debug.getJvmOption() + jmx.getJvmArgs(jiraIp) + extra
+    ): List<JvmArg> = argumentsJava8() + debug.getJvmOption() + jmx.getJvmArgs(jiraIp) + extra
+
+    fun arguments(
+        debug: JvmDebug,
+        jmx: RemoteJmx,
+        jiraIp: String,
+        jdkVersion: Int = 8
```
dagguh: We learned the hard way to avoid Kotlin default-params in API. It's better to explicitly provide the overloads.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/SetenvSh.kt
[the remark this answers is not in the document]
masiorski: looks so mych better :) I have to read sth about Kotlin

masiorski: I wanted add `val version: Int` to `JavaDevelopmentKit` in PR from bitbucket but we cannot modify public APIs. I do not want to based od `java -version`. We had a lot of problems in Jira startup scripts.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/SetenvSh.kt
masiorski: GC params are different for JDK 8 and JDK9+. For current implementation there are some defined in `JiraJvmArgs ` and `JiraGcLog `. For JDK9+ we have to add only one entry `-Xlog:gc*:file=atlassian-jira-gc-%t.log:time,uptime:filecount=5,filesize=20M`. https://dzone.com/articles/disruptive-changes-to-gc-logging-in-java-9

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/JiraJvmArgs.kt
masiorski: ok sir ;)

wyrzyk: > explicitly provide the overloads.

Why not a builder?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/SetenvSh.kt
wyrzyk: >  but we cannot modify public APIs

We can, just in a backward compatible way. Please don't misunderstand modifying and breaking. We don't **break** API without a reason.

wyrzyk: > How to expand SPI?

We can't extend the current SPI without breaking compatibility and Major release. Kotlin doesn't provide compatible overloads and migrating to Java will break Kotlin source compatibility. 

IMO there are two ways to proceed. 
1. Detect the version -> a hacky way, but transparent for the users. We can try to add a new Interface and deprecate the old one, so we'll have a way to remove the hacks in the future.
2. Add a new interface that covers Java Version. We can implement it in our JDK implementations and use to determine the version. We can treat a lack of interface implementation as Java 8 (which is behaviorally compatible).

sebapawlak: To clarify, option 1 means detecting the JDK version via ssh.

dagguh: We can future-proof ourselves from breaking Kotlin API if we add the requirement `-Xjvm-default=enable` Kotlin compiler requirement. Adding the requirement itself is breaking, but from that point on, we can use [`@JvmDefault`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/-jvm-default/index.html) to expand interfaces while preserving both Java and Kotlin compatibility at the same time. No reflection, no surprises, less tech debt, easier maintenance, less WTF factor.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/JiraJvmArgs.kt
dagguh: A builder will also work. It requires more effort, but the payoff is bigger too.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/SetenvSh.kt
wyrzyk: > No reflection, no surprises, less tech debt, easier maintenance, less WTF factor.

But we force library consumers to use a magic compiler's flag. I would not use a library that forces me to add a compiler flag to work properly.

sebapawlak: @wyrzyk could you clarify option 2? 
Do you mean adding a new Java interface **instead** of Kotlin one like this:
```
interface Jdk {
    void install(SshConnection connection);
    String use();
    String command(String options);
    Jstat getJstatMonitoring();
    String version();
}
```
```
class AdoptOpenJDK11 : JavaDevelopmentKit, Jdk {
    override fun jstatMonitoring(): Jstat = jstatMonitoring
    override fun version(): String = jdkVersion
```
or adding a new Java interface that **extends** the current one:
```
package com.atlassian.performance.tools.infrastructure.api.jvm;

interface Jdk extends JavaDevelopmentKit {
    String version();
}
```
```
class AdoptOpenJDK11 : JavaDevelopmentKit, Jdk {
    override fun version(): String = jdkVersion
```
?

How would `Jdk` be then passed to  `SetenvSh` component? I understand that still via `JiraNodeConfig` but it stores `JavaDevelopmentKit` should we try to hack and type-cast it? Or maybe add a new param in `JiraNodeConfig.Builder` with `Jdk` type and deprecate the one that accepts `JavaDevelopmentKit`?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/JiraJvmArgs.kt
```diff
@@ -13,17 +13,43 @@ class JiraJvmArgs(
...
+    ): List<JvmArg> = argumentsJava8() + debug.getJvmOption() + jmx.getJvmArgs(jiraIp) + extra
+
+    fun arguments(
+        debug: JvmDebug,
+        jmx: RemoteJmx,
+        jiraIp: String,
+        jdkVersion: Int
+    ): List<JvmArg> {
+        val jvmArg = if (jdkVersion > 8) argumentsGreaterThanJava8() else argumentsJava8()
+        return jvmArg + debug.getJvmOption() + jmx.getJvmArgs(jiraIp) + extra
```
sebapawlak: The complexity of this method has grown now. I think we could use a simple unit test here (2 test cases: Java 8 and Java 9+). @masiorski what do you think?

sebapawlak:
@masiorski that is some nice job done here. Please let us know if you need further assistance or need more guidance WRT the API compatibility.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/SetenvSh.kt
wyrzyk:
```
interface VersionedJdk extends JavaDevelopmentKit {
   Int getMajorVersion();
}
```
I'd type-cast it for backward compatibility and add a new param in the builder to clean the hacks in the future.

dagguh: > magic compiler's flag. I would not use a library that forces me to add a compiler flag to work properly.

Note it that this requirement only applies to Kotlin users, who actually implement SPI. Already they need to switch the target JVM version to 1.8 in order to implement Java SPI (if the SPI uses Java 8 default methods, e.g. `Scenario` from `jira-actions`).
This should be narrow enough, especially if we clearly document it.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/JiraJvmArgs.kt
masiorski: yea a good idea

wyrzyk:
@masiorski The last thing that left is to clean the commit history. You can squash commits like "Code refactoring" and "Fixing CI". See https://chris.beams.io/posts/git-commit/.

sebapawlak:
@masiorski could we have a separate commit for release that adds a new version in CHANGELOG.md just like here https://github.com/atlassian/infrastructure/commit/567b5c4b20ed7a0f6f03242637d0c898366e3476 pretty please.

sebapawlak:
@masiorski it is still not correct :(
In the first commit you should include the changes in the CHANGELOG but in the `unreleased` section then the second commit should move those changes to `4.11` section of the CHANGELOG and the commit message should say "Release 4.11.0". (Exactly like in already referred commit https://github.com/atlassian/infrastructure/commit/567b5c4b20ed7a0f6f03242637d0c898366e3476 )

Sorry for being so strict about this but we value the changelogs. Many times they proved to be valuable source of information for the modules so we pay attention to the details.

PR #5, opened 2019-03-22
JPERF-429: Respect `VirtualUserLoad.maxOverallLoad` when spreading lo…

dagguh:
…ad across `MulticastVirtualUsers`

Rely on `virtual-users` to slice the load. Transparently support any future sliceable VU load features.

### DIFF
--- a/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/virtualusers/MulticastVirtualUsers.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/virtualusers/MulticastVirtualUsers.kt
@@ -1,9 +1,8 @@
 package com.atlassian.performance.tools.infrastructure.api.virtualusers
 
 import com.atlassian.performance.tools.concurrency.api.submitWithLogContext
-import com.atlassian.performance.tools.virtualusers.api.VirtualUserLoad
 import com.atlassian.performance.tools.virtualusers.api.VirtualUserOptions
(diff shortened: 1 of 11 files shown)


sebapawlak:
Looks legit.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/virtualusers/MulticastVirtualUsers.kt
```diff
@@ -52,23 +51,17 @@ class MulticastVirtualUsers<out T : VirtualUsers>(
         if (nodeCount > virtualUsers) {
             throw Exception("$virtualUsers virtual users are not enough to spread into $nodeCount nodes")
         }
-        val vusPerNode = virtualUsers / nodeCount
-        val rampPerNode = load.ramp.dividedBy(nodeCount.toLong())
+        val loadSlices = load.slice(nodeCount)
```
dagguh: `slice` is covered in `virtual-users`
We could remove some unnecessary coverage from `MutlicastVirtualUsersTest`, namely the `shouldSpread*` cases
They're unit tests, so they're cheap, so it might be fine to leave them behind.
WDYT?

sebapawlak: No strong opinions on my side. It those are "just" unit tests I'm fine with leaving them behind.

PR #7, opened 2019-04-01
Issue/jperf 273 transplant ssh jira nodes

dagguh:


src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/profiler/AsyncProfiler.kt
```diff
@@ -19,7 +19,7 @@ class AsyncProfiler : Profiler {
         ssh: SshConnection,
         pid: Int
     ): RemoteMonitoringProcess {
-        ssh.execute("./async-profiler/profiler.sh -b 20000000 start $pid")
+        ssh.execute("./sync-profiler/profiler.sha -b 20000000 start $pid")
```
sebapawlak: Is this change intentional? Could you explain?

sebapawlak:
Could you explain what is the general purpose/value proposition of this PR?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/profiler/AsyncProfiler.kt
dagguh: Unintentional, thanks for catching this

dagguh:
It's a draft, I'll add more context to the commit messages when it's ready (the scope is in flux as I discover new obstacles or opportunities). But the overall goal is to ship https://ecosystem.atlassian.net/browse/JPERF-273

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/flow/install/Install.kt
```diff
@@ -0,0 +1,17 @@
...
+
+import com.atlassian.performance.tools.infrastructure.api.jira.flow.serve.Serve
+import com.atlassian.performance.tools.infrastructure.api.jira.flow.start.Start
+import com.atlassian.performance.tools.ssh.api.SshConnection
+
+/**
+ * Hooks in after installing Jira.
+ * We recommend install dependencies here rather than deferring it to later phases like [Start] or [Serve].
+ */
+interface Install {
```
pczuj: Is verb a good word for class name? What about `Installarion`/`Installer`/`Installable`?

dagguh: `InstallHook`?

pczuj: It's sill a verb if I see correctly https://en.wiktionary.org/wiki/install
Actually in case of `InstallHook`/`InstallFlow` it's already a imperative sentence which I'm also not sure if it's a good name for a class/interface.

If it's "after installing Jira", things that come to my mind:
* `JiraConfigurationTask`
* `PostJiraInstallationTask`

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/InstallableJiraIT.kt
```diff
@@ -0,0 +1,78 @@
...
+        val formula = InstallableJira(
+            name = "test",
+            jiraHomeSource = EmptyJiraHome(),
+            installHook = installHook,
+            productDistribution = PublicJiraSoftwareDistribution("7.13.0"),
+            jdk = OracleJDK()
+        )
+        val privatePort = 8080
+
+        val localReports = StandaloneUbuntu(listOf(privatePort)).runInUbuntu { ubuntu ->
```
dagguh: Depends on https://github.com/atlassian/ssh-ubuntu/pull/5

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/AdoptOpenJdk11IT.kt
```diff
@@ -6,7 +6,7 @@ class AdoptOpenJdk11IT {
 
     @Test
     fun shouldSupportJstat() {
-        val expectedJstatHeader = "Timestamp         S0     S1     E      O      M     CCS    YGC     YGCT    FGC    FGCT    CGC    CGCT     GCT   "
+        val expectedJstatHeader = "Timestamp         S0     S1     E      EmptyJiraHome      M     CCS    YGC     YGCT    FGC    FGCT    CGC    CGCT     GCT   "
```
dagguh: IDEA 🤦‍♂️

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/MySqlDatabase.kt
```diff
@@ -61,4 +71,50 @@ class MySqlDatabase(
...
+    }
+}
+
+private class MysqlJdbc(
+    private val databaseIp: String
+) : Install {
+    override fun install(ssh: SshConnection, jira: InstalledJira): Start {
+        Sed().replace(
+            connection = ssh,
+            expression = "(<url>.*(@(//)?|//))" + "([^:/]+)" + "(.*</url>)",
```
dagguh: It still preserves the old assumption that the dbconfig.xml from Jira home is set up for MySQL already and all we need to do is to switch the IP. We can make it more robust in another PR.
Here's how the assumed `dbconfig.xml` looks like:
```xml
<?xml version="1.0" encoding="UTF-8"?>

<jira-database-config>
  <name>defaultDS</name>
  <delegator-name>default</delegator-name>
  <database-type>mysql</database-type>
  <jdbc-datasource>
    <url>jdbc:mysql://18.184.122.128:3306/jiradb?useUnicode=true&amp;characterEncoding=UTF8&amp;sessionVariables=storage_engine=InnoDB</url>
    <driver-class>com.mysql.jdbc.Driver</driver-class>
    <username>jirauser</username>
    <password>jirauser</password>
    <pool-min-size>20</pool-min-size>
    <pool-max-size>20</pool-max-size>
    <pool-max-wait>30000</pool-max-wait>
    <validation-query>select 1</validation-query>
    <min-evictable-idle-time-millis>60000</min-evictable-idle-time-millis>
    <time-between-eviction-runs-millis>300000</time-between-eviction-runs-millis>
    <pool-max-idle>20</pool-max-idle>
    <pool-remove-abandoned>true</pool-remove-abandoned>
    <pool-remove-abandoned-timeout>300</pool-remove-abandoned-timeout>
    <pool-test-while-idle>true</pool-test-while-idle>
    <validation-query-timeout>3</validation-query-timeout>
  </jdbc-datasource>
</jira-database-config>
```

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/DataCenterNodeFormula.kt
```diff
@@ -0,0 +1,36 @@
...
+//    override fun provision(): StoppedNode {
+//
+//        val provisionedNode = base.provision()
+//        val localSharedHome = sharedHome.get().localSharedHome
+//
+//        provisionedNode.ssh.newConnection().use {
+//            sharedHome.get().mount(it)
+//            val jiraHome = provisionedNode.jiraHome
+//
+//            it.execute("echo ehcache.object.port = 40011 >> $jiraHome/cluster.properties")
```
dagguh: This stuff looks like another install hook. TBD later in this PR

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/flow/install/DefaultJiraInstallation.kt
```diff
@@ -0,0 +1,27 @@
...
+import com.atlassian.performance.tools.infrastructure.api.distribution.ProductDistribution
+import com.atlassian.performance.tools.infrastructure.api.jira.JiraHomeSource
+import com.atlassian.performance.tools.infrastructure.api.jira.flow.InstalledJira
+import com.atlassian.performance.tools.infrastructure.api.jira.flow.TcpServer
+import com.atlassian.performance.tools.infrastructure.api.jira.flow.report.ReportTrack
+import com.atlassian.performance.tools.infrastructure.api.jvm.JavaDevelopmentKit
+import com.atlassian.performance.tools.ssh.api.SshConnection
+
+class DefaultJiraInstallation(
+    private val jiraHomeSource: JiraHomeSource, // TODO add database too, but how to handle databaseIp? Perhaps as a wrapper, which could be applied by the "DB installing" code?
```
dagguh: > TODO add database too, but how to handle databaseIp? Perhaps as a wrapper, which could be applied by the "DB installing" code?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/flow/start/DefaultPostStartHook.kt
```diff
@@ -0,0 +1,20 @@
...
+
+    override fun hook(
+        ssh: SshConnection,
+        jira: StartedJira,
+        track: ReportTrack
+    ) {
+        listOf(
+            JiraLogs(),
+            JstatHook()
+            //,RestUpgrade(config.launchTimeouts, "admin", "admin") TODO requires database to work
```
dagguh: > RestUpgrade(config.launchTimeouts, "admin", "admin") TODO requires database to work

dagguh: Can we control the order of things happening before or after the upgrade?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/flow/install/UbuntuSysstat.kt
```diff
@@ -0,0 +1,38 @@
...
+import com.atlassian.performance.tools.infrastructure.api.os.OsMetric
+import com.atlassian.performance.tools.infrastructure.api.os.Ubuntu
+import com.atlassian.performance.tools.infrastructure.jira.flow.RemoteMonitoringProcessReport
+import com.atlassian.performance.tools.ssh.api.SshConnection
+
+class UbuntuSysstat : PostInstallHook {
+
+    override fun hook(
+        ssh: SshConnection,
+        jira: InstalledJira,
```
dagguh: Design flaw: we require `InstalledJira`, but it is not used. E.g. I should be able to run sysstat before installing Jira.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/flow/install/UbuntuSysstat.kt
```diff
@@ -0,0 +1,38 @@
...
+        flow: JiraNodeFlow
+    ) {
+        val postStartHooks = Ubuntu()
+            .metrics(ssh)
+            .map { InstalledOsMetric(it) }
+        flow.postStartHooks.addAll(postStartHooks)
```
dagguh: Design flaw: we hardcode when the metrics will be started. If I want to start them before Jira starts, I gotta write an entire new `UbuntuSysstatHook`.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/flow/install/JiraLogs.kt
```diff
@@ -0,0 +1,15 @@
...
+
+import com.atlassian.performance.tools.infrastructure.api.jira.flow.JiraNodeFlow
+import com.atlassian.performance.tools.infrastructure.api.jira.flow.report.StaticReport
+import com.atlassian.performance.tools.ssh.api.SshConnection
+
+class JiraLogs : InstalledJiraHook {
+
+    override fun hook(ssh: SshConnection, jira: InstalledJira, flow: JiraNodeFlow) {
+        listOf(
+            StaticReport("${jira.home}/log/atlassian-jira.log"),
```
dagguh: ### Require more: `StartedJira`

We expect these logs to appear after starting Jira, so we can formalize this behavioral pre-condition by `JiraLogs: StartedJiraHook`.

### Require less: `InstalledJira`
However, they might be there before start as well (e.g. if we started and stopped the node before) and technically the code only needs to `JiraLogs : InstalledJiraHook`.

### Which one?
Should we require more or less? If more, then it's harder to make a mistake. If less, then it covers less use cases.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/flow/install/UbuntuSysstat.kt
dagguh: Fixed in 4bfa05 by transparently translating low-requierement hooks to high-requirement hooks.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/flow/install/DataCenterHook.kt
```diff
@@ -0,0 +1,24 @@
...
+) : InstalledJiraHook {
+
+    override fun hook(
+        ssh: SshConnection,
+        jira: InstalledJira,
+        flow: JiraNodeFlow
+    ) {
+        val localSharedHome = sharedHome.localSharedHome
+        sharedHome.mount(ssh)
+        val jiraHome = jira.home  // TODO what's the difference between localSharedHome and jiraHome? should both be hookable?
```
pczuj: Can't see a difference between "jirahome" and "jira-local-home"

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/flow/install/JiraLogs.kt
pczuj: > make a mistake
What is the mistake and what are the consequences of it?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/flow/install/DefaultJiraInstallation.kt
```diff
@@ -0,0 +1,26 @@
...
+    private val jdk: JavaDevelopmentKit
+) : JiraInstallation {
+
+    override fun install(
+        ssh: SshConnection,
+        server: TcpServer,
+        flow: JiraNodeFlow
+    ): InstalledJira {
+        val installation = productDistribution.install(ssh, ".")
+        val home = jiraHomeSource.download(ssh)
```
pczuj: Just a note: I'd remove the knowledge of jirahome being downloaded from `JiraHomeSource` and name it `#install()`. The `JiraHomeSource` could in the future be implemented as offline jirahome generator.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/flow/install/JiraHomeProperty.kt
```diff
@@ -0,0 +1,16 @@
+package com.atlassian.performance.tools.infrastructure.api.jira.flow.install
+
+import com.atlassian.performance.tools.infrastructure.api.jira.flow.JiraNodeFlow
+import com.atlassian.performance.tools.ssh.api.SshConnection
+
+class JiraHomeProperty : InstalledJiraHook {
```
pczuj: This could be part of jirahome installation

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/flow/install/JiraInstallation.kt
```diff
@@ -0,0 +1,14 @@
+package com.atlassian.performance.tools.infrastructure.api.jira.flow.install
+
+import com.atlassian.performance.tools.infrastructure.api.jira.flow.TcpServer
+import com.atlassian.performance.tools.infrastructure.api.jira.flow.JiraNodeFlow
+import com.atlassian.performance.tools.ssh.api.SshConnection
+
+interface JiraInstallation {
```
pczuj: Funny that this is also indirectly `TcpServerHook` (logicly), but the `TcpServerHook` is also `InstalledJiraHook`, so basically you could install Jira on top of tcp server with Jira, which makes sense.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/flow/install/DataCenterHook.kt
```diff
@@ -0,0 +1,24 @@
+package com.atlassian.performance.tools.infrastructure.api.jira.flow.install
+
+import com.atlassian.performance.tools.infrastructure.api.jira.SharedHome
+import com.atlassian.performance.tools.infrastructure.api.jira.flow.JiraNodeFlow
+import com.atlassian.performance.tools.ssh.api.SshConnection
+
+class DataCenterHook(
```
pczuj: I'd name it differently, like e.g. `DataCenterUpgrade` or `DataCenterInstallation`

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/flow/server/SystemLog.kt
```diff
@@ -0,0 +1,12 @@
+package com.atlassian.performance.tools.infrastructure.api.jira.flow.server
+
+import com.atlassian.performance.tools.infrastructure.api.jira.flow.report.Report
+import com.atlassian.performance.tools.infrastructure.api.jira.flow.report.StaticReport
+import com.atlassian.performance.tools.ssh.api.SshConnection
+
+class SystemLog : Report {
```
pczuj:
```
class SystemLog : Report by StaticReport("/var/log/syslog")
```

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/flow/start/DefaultStartedJiraHook.kt
```diff
@@ -0,0 +1,21 @@
...
+
+    override fun hook(
+        ssh: SshConnection,
+        jira: StartedJira,
+        flow: JiraNodeFlow
+    ) {
+        listOf(
+            JiraLogs(),
+            JstatHook()
+            //,RestUpgrade(config.launchTimeouts, "admin", "admin") TODO requires database to work
```
pczuj: Do you mean credentials? This knowledge from database could be bypassed if the user would always be admin/admin (or any defined in the test config). 
Following KB could be the material for another "Hook" that would ensure that: https://confluence.atlassian.com/jira/retrieving-the-jira-administrator-192836.html
We would need to have an sql client that could work with every database Jira supports - possibly some kind of application to reset jira password.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/flow/TcpServer.kt
```diff
@@ -0,0 +1,19 @@
+package com.atlassian.performance.tools.infrastructure.api.jira.flow
+
+class TcpServer(
```
pczuj: Should this be `TcpServer` or should it be `UbuntuServer` with ability to use it's shell? I feel like we are already very heavily depending on Ubuntu, maybe we should be explicit and honest? also having `Ubuntu` or `Bash` (or any shell) abstraction would make Hooks do less gambling around what can they do via the SshConnection (BTW the SSH should also be hidden abstraction, because no Hook actually cares about the shell being accessed remotely - they care about executing things via Ubuntu shell).

dagguh:
FYI I'm doing this on my 20% time, so expect slow progress

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/flow/TcpServer.kt
dagguh: This is expliciltly a TCP server, which is OS-agnostic. It doesn't know SSH.
Examples of a TCP server:
* Tomcat in Jira
* MySQL server
* Chromedriver W3C WD API server

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/InstallableDatabase.kt
```diff
@@ -0,0 +1,8 @@
+package com.atlassian.performance.tools.infrastructure.api.database
+
+import com.atlassian.performance.tools.infrastructure.api.jira.flow.install.Install
+
+interface InstallableDatabase : Database {
+
+    fun installInJira(databaseIp: String): Install
```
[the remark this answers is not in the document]
dagguh: Because the callers don't know the JDBC URL. `aws-infrastructure` only generates the IP.
Perhaps this is the same smell as in https://github.com/atlassian/aws-infrastructure/pull/15/files#r273025547 : we're designing low-level API to match the existing caller patterns.
Perhaps it's the `aws-infrastructure`, who should ask for the `fun install(databaseIp: String): Install`

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/flow/JiraNodeFlow.kt
```diff
@@ -0,0 +1,62 @@
...
+import java.util.concurrent.ConcurrentLinkedQueue
+
+@ThreadSafe
+class JiraNodeFlow {
+
+    private val preInstallHooks: Queue<TcpServerHook> = ConcurrentLinkedQueue()
+    private val postInstallHooks: Queue<InstalledJiraHook> = ConcurrentLinkedQueue()
+    private val preStartHooks: Queue<InstalledJiraHook> = ConcurrentLinkedQueue()
+    private val postStartHooks: Queue<StartedJiraHook> = ConcurrentLinkedQueue()
+    val reports: Queue<Report> = ConcurrentLinkedQueue()
```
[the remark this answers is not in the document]
wyrzyk: Do we need both to be public? Isn't flatMap an internal call?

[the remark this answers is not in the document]
dagguh: No need. We can hide it 👍

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/flow/start/DefaultStartedJiraHook.kt
dagguh: I meant that `RestUpgrade` has a hidden dependency on DB. All other hooks only depended on what's injected in their constructors or method params. I guess the REST upgrade hook could become an injectable part of the DB hook.
If that happens then the credentials are an internal contract of the hook, so that's all good

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/ThreadDump.kt
```diff
@@ -26,7 +26,7 @@ class ThreadDump(
      */
     fun gather(connection: SshConnection, destination: String) {
         val threadDumpName = Instant.now().toEpochMilli()
-        val command = "${jdk.use()}; jcmd $pid Thread.print > $destination/${threadDumpName}"
+        val command = "${jdk.use()}; jcmd $pid Thread.print > $destination/$threadDumpName"
```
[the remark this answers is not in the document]
dagguh: Ok, I'll try to make this PR smaller. It's too big to comprehend.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/hook/JiraNodeHooks.kt
```diff
@@ -0,0 +1,25 @@
+package com.atlassian.performance.tools.infrastructure.api.jira.hook
+
+import com.atlassian.performance.tools.infrastructure.api.jira.JiraNodeConfig
+import com.atlassian.performance.tools.infrastructure.api.jira.hook.install.DefaultPostInstallHook
+import com.atlassian.performance.tools.infrastructure.api.jira.hook.install.PreInstallHooks
+import com.atlassian.performance.tools.infrastructure.api.jira.hook.start.DefaultPostStartHook
+import net.jcip.annotations.ThreadSafe
+
+@ThreadSafe
+class JiraNodeHooks {
```
dagguh, line 10: Remove this "entry point"? It can be replaced by `PreInstallHooks` + factory methods

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/hook/JiraNodeHooks.kt
```diff
@@ -0,0 +1,25 @@
...
+
+    val preInstall = PreInstallHooks()
+    val postInstall = preInstall.postInstall
+    val preStart = postInstall.preStart
+    val postStart = preStart.postStart
+    val reports = postStart.reports
```
dagguh, line 16: These are all already available via `preInstall`. They serve as syntax sugar to avoid long chains.
Before:
```kotlin
hooks.preInstall.postInstall.preStart.postStart.reports.list()
```

After:
```
hooks.reports.list()
```

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/hook/install/AsyncProfilerHook.kt
```diff
@@ -0,0 +1,55 @@
+package com.atlassian.performance.tools.infrastructure.api.jira.hook.install
+
+import com.atlassian.performance.tools.infrastructure.api.jira.hook.start.PostStartHooks
+import com.atlassian.performance.tools.infrastructure.api.jira.hook.report.Report
+import com.atlassian.performance.tools.infrastructure.api.jira.hook.start.PostStartHook
+import com.atlassian.performance.tools.ssh.api.SshConnection
+import java.net.URI
+
+class AsyncProfilerHook : PreInstallHook {
```
dagguh, line 9: @mzyromski-atlassian This is one of the more advanced usage of hooks.
It installs a profiler before Jira is installed. Then starts the profiler after Jira starts (using PID provided by hooks). And it registers a flamegraph file for download.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/DatabaseIpConfig.kt
```diff
@@ -0,0 +1,25 @@
...
+
+import com.atlassian.performance.tools.infrastructure.api.Sed
+import com.atlassian.performance.tools.infrastructure.api.jira.flow.JiraNodeFlow
+import com.atlassian.performance.tools.infrastructure.api.jira.flow.install.InstalledJira
+import com.atlassian.performance.tools.infrastructure.api.jira.flow.install.InstalledJiraHook
+import com.atlassian.performance.tools.ssh.api.SshConnection
+
+class DatabaseIpConfig(
+    private val databaseIp: String
+) : InstalledJiraHook {
```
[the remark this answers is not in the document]
dagguh: - [x] KDoc for `interface`s
- [ ] KDoc for `class`es

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/flow/install/UbuntuSysstat.kt
dagguh: I tried to `fun deferPostStart(preInstall: PreInstallHook): PreStartHook`. But `PreStartHook` cannot hook itself into the past (pre-install, post-install), while the original `preInstall` could. So the compiler prohibited me from implementing such a generic deferall. So you gotta write a new hook. See the difference between `EarlyUbuntuSysstat` vs `LateUbuntuSysstat`.

dagguh:
@mzyromski-atlassian this is too big, so I started splitting it.
Here are the chunks:
* #100

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/flow/install/JiraLogs.kt
dagguh: The mistake would be you hook `JiraLogs`, but the download would fail if you do it before Jira is started. But we can make the download more forgiving. Also the start/stop/start case is kinda forced. Also the API consumer can inject a different way of downloading the log even if they have an exotic use case. So I'll just make it work for the most popular case - getting the logs after Jira was started.

PR #9, opened 2019-04-04
Mark release 4.12.1

wyrzyk:


dagguh:
Build failed, but we don't have the artifacts so we can't see the details 😞 
We should either add artifacts to TravisCI builds (our own S3 bucket, optional for contributors) or switch to CircleCI.

PR #12, opened 2019-04-11
Fix chopped multiline in Travis YAML

dagguh:
https://travis-ci.com/atlassian/infrastructure/jobs/192189041/config failed and did not notify on Slack.
I used http://www.yamllint.com/ to parse the YAML and it turns out the current multiline syntax adds spaces on every new line. I used the same linter to check the fixed version.


dagguh:
PS. Boo for Travis for silently ignoring an invalid Slack config.

PR #13, opened 2019-04-17
JPERF-460 Add more patience to `lftp` installation

wyrzyk:


src/main/kotlin/com/atlassian/performance/tools/infrastructure/HttpResource.kt
```diff
@@ -25,7 +25,7 @@ class HttpResource(
         destination: String,
         timeout: Duration
     ) {
-        Ubuntu().install(ssh, listOf("lftp"))
+        Ubuntu().install(ssh, listOf("lftp"), Duration.ofMinutes(2))
```
dagguh: The CI artifacts do not show it was a timeout. `IdempotentAction` logs the stacktraces for each attempt on debug level, which seems to be missing from the test stdout. We should either start logging it or make `IdempotentAction` link the cause (first? last? `addSuppressed()` all?) in the final exception. 
Gotta have a red test, which **shows** the problem, which justifies the fix.

CHANGELOG.md
```diff
@@ -26,6 +26,11 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/infrastructure/compare/release-4.12.1...master
 
+### Fixed
+- Add more patience to `lftp` installation. Fix [JPERF-460]
```
dagguh: Nit:
```suggestion
- Have more patience for `lftp` installation. Fix [JPERF-460].
```

CHANGELOG.md
```diff
@@ -26,6 +26,11 @@ Dropping a requirement of a major version of a dependency is a new contract.
...
+
+[JPERF-460]: https://ecosystem.atlassian.net/browse/JPERF-460
+
 ## [4.12.1] - 2019-04-04
```
dagguh: This new fix is marked as being after 4.12.1 (which supposedly was released on April 4), but [that version currently does not exist](https://packages.atlassian.com/webapp/#/search/package/eyJzZWFyY2giOiJnYXZjIiwiZ3JvdXBJRCI6ImNvbS5hdGxhc3NpYW4ucGVyZm9ybWFuY2UudG9vbHMiLCJhcnRpZmFjdElEIjoiaW5mcmFzdHJ1Y3R1cmUiLCJ2ZXJzaW9uIjoiNC4xMi4xIiwic2VsZWN0ZWRQYWNrYWdlVHlwZSI6eyJpZCI6ImdhdmMiLCJkaXNwbGF5TmFtZSI6Ik1hdmVuIEdBVkMiLCJpY29uIjoicG9tIn19). This is the disadvantage of marking releases before they actually happen.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/HttpResource.kt
wyrzyk: > The CI artifacts do not show it was a timeout. `IdempotentAction` logs the stacktraces for each attempt on debug level, which seems to be missing from the test stdout. We should either start logging it or make `IdempotentAction` link the cause (first? last? `addSuppressed()` all?) in the final exception.
> Gotta have a red test, which **shows** the problem, which justifies the fix.

I run the test locally multiple times, and it was failing due to timeout.

dagguh: Aight, raised https://ecosystem.atlassian.net/browse/JPERF-461

PS. no need to quote-reply in threads, it can make the convos less readable

PR #16, opened 2019-04-23
Issue/jperf 468 thread safe ubuntu

wyrzyk:


src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/os/UbuntuIT.kt
```diff
@@ -0,0 +1,54 @@
...
+import com.atlassian.performance.tools.sshubuntu.api.SshUbuntuContainer
+import org.junit.After
+import org.junit.Before
+import org.junit.Test
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.Executors
+import java.util.concurrent.TimeUnit
+
+class UbuntuIT {
```
dagguh: Tested it locally and it's indeed red-green
Neat :)

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/os/Ubuntu.kt
```diff
@@ -4,11 +4,17 @@ import com.atlassian.performance.tools.infrastructure.Iostat
...
 import com.atlassian.performance.tools.ssh.api.SshConnection
+import net.jcip.annotations.ThreadSafe
 import org.apache.logging.log4j.Level
 import java.time.Duration
 
+@ThreadSafe
 class Ubuntu {
 
+    private companion object {
+        private val lock = Object()
```
dagguh: This lock is global, which means that all Ubuntu installs on all infrastructure (even on separate instances or even networks) will be serialized. If tests use a lot of parallelization, this leads to very slow (e.g. 1h 20m penalty for a single apt-get call), infrastructure setup, timeouts, SSH connection deaths and AWS resource lifespan expiry.
Raised https://ecosystem.atlassian.net/browse/JPERF-478

PR #18, opened 2019-04-29
JC-316 Checking checksum before unzip

lipinskipawel:
Unable to provision the JIRA node due to file corruption.
Should we compare only the checksum (first number) or all the output from the cksum tool?

https://ecosystem.atlassian.net/browse/JPERF-474

### DIFF
--- a/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/OracleJDK.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/OracleJDK.kt
@@ -13,6 +13,7 @@ class OracleJDK : VersionedJavaDevelopmentKit {
     private val jdkUrl = URI.create("https://download.oracle.com/otn-pub/java/jdk/8u$jdkUpdate-b11/d54c1d3a095b4ff2b6607d096fa80163/$jdkArchive")
     private val jdkBin = "~/jdk1.8.0_$jdkUpdate/jre/bin/"
     private val bin = "~/jdk1.8.0_$jdkUpdate/bin/"
+    private val CHECKSUM_FROM_ORACLE_JDK_TAR_GZ_OUTPUT = "19878902 185540433 jdk-8u131-linux-x64.tar.gz"
     override val jstatMonitoring = Jstat(bin)
 
     @Deprecated(
@@ -25,8 +26,20 @@ class OracleJDK : VersionedJavaDevelopmentKit {
 
     override fun install(connection: SshConnection) {
         download(connection)
-        connection.execute("tar -xzf $jdkArchive")
-        connection.execute("echo '${use()}' >> ~/.bashrc")
+        val execute = connection.execute("cksum jdk-8u131-linux-x64.tar.gz")
+
+        // Checks 3 times if the downloaded archive of JDK has a proper checksum.
+        // This is done in order to prevent failure to execute unzipping a corrupted file.
+        for (attempt in 0..2) {
+            if (execute.output.trim() == CHECKSUM_FROM_ORACLE_JDK_TAR_GZ_OUTPUT) {
+                connection.execute("tar -xzf $jdkArchive")
+                connection.execute("echo '${use()}' >> ~/.bashrc")
+                break
+            } else {
+                connection.execute("rm -rf jdk-8u131-linux-x64.tar.gz")
+                download(connection)
+            }
+        }
     }
 
     override fun use(): String = "export PATH=$jdkBin:$bin:${'$'}PATH"
@@ -51,3 +64,4 @@ class OracleJDK : VersionedJavaDevelopmentKit {
         }
     }
 }
+


dagguh:
I can't find JC-316 on a public issue tracker. I'd raise a JPERF, as linked in the README.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/OracleJDK.kt
```diff
@@ -25,8 +26,20 @@ class OracleJDK : VersionedJavaDevelopmentKit {
...
+        val execute = connection.execute("cksum jdk-8u131-linux-x64.tar.gz")
+
+        // Checks 3 times if the downloaded archive of JDK has a proper checksum.
+        // This is done in order to prevent failure to execute unzipping a corrupted file.
+        for (attempt in 0..2) {
+            if (execute.output.trim() == CHECKSUM_FROM_ORACLE_JDK_TAR_GZ_OUTPUT) {
+                connection.execute("tar -xzf $jdkArchive")
+                connection.execute("echo '${use()}' >> ~/.bashrc")
+                break
+            } else {
```
dagguh: Why can it be downloaded with a wrong checksum? Is it caused by our transport or is it it randomly corrupted by the `download.oracle.com` server?
Is it because our `download` is also retrying stuff and one call can overwrite the other? If so, we could download each attempt into a different target file and avoid the corruption?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/OracleJDK.kt
```diff
@@ -25,8 +26,20 @@ class OracleJDK : VersionedJavaDevelopmentKit {
 
     override fun install(connection: SshConnection) {
         download(connection)
-        connection.execute("tar -xzf $jdkArchive")
-        connection.execute("echo '${use()}' >> ~/.bashrc")
+        val execute = connection.execute("cksum jdk-8u131-linux-x64.tar.gz")
+
+        // Checks 3 times if the downloaded archive of JDK has a proper checksum.
```
dagguh: This assumes there's a problem and assumes this will fix it. A test would go a long way towards validating both these assumptions. We already have a few integration tests which test the download/installation process of various software, e.g. `Chromium69IT`.
Pro-tip: if you introduce the test in one commit and the fix in the next one, you'll:
* show that the test faithfully represents a real world problem via red CI
* show that the fix actually makes a difference via green CI
* enable exploring different fixes for the same problem

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/OracleJDK.kt
lipinskipawel: I'm not sure which step causes the corruption of a file, maybe it is by transport or it is something wrong with download.oracle.com server.
Either way, I think that our download method is working properly and can overwrite an existing file.

dagguh: Note that it sends an SSH command. If it times out on our end, it can still be running on the remote system. This way, with retries, you can have multiple concurrent processes writing to the same file, causing the corruption.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/OracleJDK.kt
sebapawlak: Inline code comment is generally at smell it indicated that the code is unreadable and not self-explanatory.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/OracleJDK.kt
```diff
@@ -25,8 +26,20 @@ class OracleJDK : VersionedJavaDevelopmentKit {
...
+        // This is done in order to prevent failure to execute unzipping a corrupted file.
+        for (attempt in 0..2) {
+            if (execute.output.trim() == CHECKSUM_FROM_ORACLE_JDK_TAR_GZ_OUTPUT) {
+                connection.execute("tar -xzf $jdkArchive")
+                connection.execute("echo '${use()}' >> ~/.bashrc")
+                break
+            } else {
+                connection.execute("rm -rf jdk-8u131-linux-x64.tar.gz")
+                download(connection)
```
sebapawlak: Does it mean we are downloading up to 9 times (3 times 3)?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/OracleJDK.kt
```diff
@@ -13,6 +13,7 @@ class OracleJDK : VersionedJavaDevelopmentKit {
     private val jdkUrl = URI.create("https://download.oracle.com/otn-pub/java/jdk/8u$jdkUpdate-b11/d54c1d3a095b4ff2b6607d096fa80163/$jdkArchive")
     private val jdkBin = "~/jdk1.8.0_$jdkUpdate/jre/bin/"
     private val bin = "~/jdk1.8.0_$jdkUpdate/bin/"
+    private val CHECKSUM_FROM_ORACLE_JDK_TAR_GZ_OUTPUT = "19878902 185540433 jdk-8u131-linux-x64.tar.gz"
```
sebapawlak: We definitely need an integration test for that. From now on upgrading JDK will no longer as easy as bumping `jdkUpdate` but also the checksum. The test would verify that.

Also as you can see here we use camelCase for `private val`s rather than SNAKE_CASE. It even looks a bit weird.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/OracleJDK.kt
sebapawlak: Actually it is not guaranteed that the `curl` will yield z `tar.gz`. Try making a typo in the URL (just change some letter) and run the command locally. The downloaded file will be still named `something.jdk.tar.gz` but it will be a HTML page actually with 404 error. So maybe we are trying to uncompress a HTML page served by the Oralce's server and perhaps the message on the page is meaningful.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/OracleJDK.kt
```diff
@@ -25,8 +26,20 @@ class OracleJDK : VersionedJavaDevelopmentKit {
 
     override fun install(connection: SshConnection) {
         download(connection)
-        connection.execute("tar -xzf $jdkArchive")
-        connection.execute("echo '${use()}' >> ~/.bashrc")
+        val execute = connection.execute("cksum jdk-8u131-linux-x64.tar.gz")
+
+        // Checks 3 times if the downloaded archive of JDK has a proper checksum.
+        // This is done in order to prevent failure to execute unzipping a corrupted file.
+        for (attempt in 0..2) {
```
sebapawlak: I think this attempts should be done as a part of download method. We would keep single retry loop in such case making the code much more readable.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/OracleJDK.kt
```diff
@@ -25,8 +26,20 @@ class OracleJDK : VersionedJavaDevelopmentKit {
...
+            if (execute.output.trim() == CHECKSUM_FROM_ORACLE_JDK_TAR_GZ_OUTPUT) {
+                connection.execute("tar -xzf $jdkArchive")
+                connection.execute("echo '${use()}' >> ~/.bashrc")
+                break
+            } else {
+                connection.execute("rm -rf jdk-8u131-linux-x64.tar.gz")
```
sebapawlak: This command will stop working if we change the version of the jdk, please reuse the variables.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/OracleJDK.kt
sebapawlak: Let's name it `jdkArchiveChecksum`.

sebapawlak:
Please add a `CHANGELOD.md` entry for this changes.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/OracleJDK.kt
```diff
@@ -25,8 +26,20 @@ class OracleJDK : VersionedJavaDevelopmentKit {
...
     override fun install(connection: SshConnection) {
         download(connection)
-        connection.execute("tar -xzf $jdkArchive")
-        connection.execute("echo '${use()}' >> ~/.bashrc")
+        val execute = connection.execute("cksum jdk-8u131-linux-x64.tar.gz")
+
+        // Checks 3 times if the downloaded archive of JDK has a proper checksum.
+        // This is done in order to prevent failure to execute unzipping a corrupted file.
+        for (attempt in 0..2) {
+            if (execute.output.trim() == CHECKSUM_FROM_ORACLE_JDK_TAR_GZ_OUTPUT) {
```
sebapawlak: Also if it fails for the third time I would love to see some logging and an `Exception` thrown. Right now we are ignoring the problem. In this case it is definitely better to fail fast/early with a clear cause provided.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/OracleJDK.kt
wyrzyk: @sebapawlak  We already have a test that covers downloading the resource. It'll fail if the checksum is not valid.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/OracleJDK.kt
wyrzyk: I tried to reproduce the issue by running two downloads at the same time. Always the one which started second succeeds and the first fails. 

I also investigated failed build and there were no retries. The file has been downloaded. It has 177mb, but I'm unable to unzip it (I've tried on remote Ubuntu machine and on my mac).

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/OracleJDK.kt
dagguh: PS. `IdempotentAction` already covers retries and logging

lipinskipawel:
This change is resolved by https://ecosystem.atlassian.net/browse/JPERF-474. No need for code changes.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/OracleJDK.kt
dagguh: > The downloaded file will be still named something.jdk.tar.gz but it will be a HTML page actually with 404 error.

I suspect this the most. We'd have to improve our `curl`-fu to make it smarter and learn what's the error. E.g. maybe the cookie hack is no longer valid.

PR #19, opened 2019-05-02
Postgres db support rebased 2

dagguh:


dagguh:
Oops, meant it to target like this: https://github.com/kema2000/infrastructure/pull/2

PR #22, opened 2019-06-21
Issue/jperf 496 fix dataset download

wyrzyk:


src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/dataset/HttpDatasetPackageIT.kt
```diff
@@ -17,11 +21,43 @@ class HttpDatasetPackageIT {
         )
 
         val unpackedPath = SshUbuntuContainer().start().use { sshUbuntu ->
-            return@use sshUbuntu.toSsh().newConnection().use { connection ->
-                return@use dataset.download(connection)
+            val ssh = sshUbuntu.toSsh()
+            return@use RandomFilesGenerator(ssh).start().use {
```
dagguh: Clever :)

dagguh: TravisCI doesn't let us see the fail details, so I ran it locally and indeed it is the same symptom I saw on production:
```
java.lang.Exception: Expected one new folder. Found [01b0f471-6fc0-4fca-9639-efb9008e5b8d, 01db8de6-1aab-4cca-8111-4ada812f6dee, 
```

Good repro

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/dataset/HttpDatasetPackageIT.kt
```diff
@@ -17,11 +21,43 @@ class HttpDatasetPackageIT {
...
+
+                fun start() {
+                    executor.execute {
+                        while (createNewFiles.get()) {
+                            ssh.newConnection().use {
+                                it.safeExecute(
+                                    cmd = "touch ${UUID.randomUUID()}"
+                                )
+                            }
+                            Thread.sleep(20)
```
dagguh: This generates files very quickly. The full assertion error message is very long. Also, the stdout is very spammy.
When I increased this to `Thread.sleep(2000)`, it still reproduced:
```
java.lang.Exception: Expected one new folder. Found [0af677fb-c52d-400d-9711-c45318229679, 24f72cb4-d041-4362-9c55-46a4b4352ade, 6ad6f223-698e-43d3-a628-97fddbf91979, 850f99e7-3b63-47be-8823-ecac1c7ddba5, aa078242-bd1a-4ec4-bed3-c6ae35a0ade0, database].
```

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/dataset/FileArchiver.kt
```diff
@@ -22,6 +22,22 @@ class FileArchiver {
         }
     }
 
+    /**
+     *  Decompress `tar.bz2` archives to the current directory.
+     *
+     *  @return files processed
+     */
+    fun verboseUnzip(
```
dagguh: Does it have to be public? If so, we gotta mark it as `Added` in the changelog.

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/dataset/HttpDatasetPackageIT.kt
wyrzyk: Maybe it depends on the internet connection. I had 2 files generated for 20ms interval.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/dataset/FileArchiver.kt
wyrzyk: Can be internal.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/dataset/HttpDatasetPackage.kt
```diff
@@ -39,23 +39,10 @@ internal class HttpDatasetPackage(
...
-    ): Set<String> {
-        return ssh
-            .execute("ls")
-            .output
-            .split("\\s".toRegex())
-            .filter { it.isNotBlank() }
-            .toSet()
+        return FileArchiver()
+            .verboseUnzip(ssh, resourceName, timeForUnzipping)
+            .first()
```
bogdanb: Doesn’t this lose the check that it contains exactly one folder? Not sure how valuable that was.

wyrzyk: It loses the check, but I don't know if it was valuable.

dagguh: If we want to get the check back we can go from `first` to `single`

wyrzyk: Verbose tar returns a list of all files extracted. It'll be never `single`.

dagguh: All dataset packages are reduced to a single directory (see `DatasetPackage.download`).
We don't have a mechanism for choosing a directory if multiple would be presented.

bogdanb: I was thinking more of giving a meaningful error message early to users who might be trying to customize parts of the process and mess up somewhere. (Ahem...)

dagguh: @wyrzyk I see you pushed the check
Here's my take: https://github.com/atlassian/infrastructure/commit/e42029eb7c48036ed89dc4fcfccb2142c64bd24b
It has less moving parts and includes a more meaningful error message

wyrzyk: @dagguh It may be premature optimisation, but having in mind that our datasets can contain lots of files (attachments?), I didn't want to hold/allocate more objects than needed.

dagguh: Right, we can mix in the `.asSequence()` there
The `Set` holds only the top-level paths

wyrzyk: @dagguh I'll merge the changes as is. Feel free to patch it in the next PR.

wyrzyk: @bogdanb thanks for spotting the problem (the actual issue and the implementation flaw).

PR #26, opened 2019-06-25
JPERF-504: Fix dataset unzipping performance.

wyrzyk:
The issue was introduced while working on JPERF-496.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,11 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/infrastructure/compare/release-4.12.4...master
 
+### Fixed
+- Fix dataset unzipping performance. Fix [JPERF-504].
+
+[JPERF-504]: https://ecosystem.atlassian.net/browse/JPERF-504
+
 ## [4.12.4] - 2019-06-25
 [4.12.4]: https://github.com/atlassian/infrastructure/compare/release-4.12.3...release-4.12.4
 
--- a/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/dataset/FileArchiver.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/dataset/FileArchiver.kt
@@ -23,18 +23,19 @@ class FileArchiver {
     }
 
     /**
-     *  Decompress `tar.bz2` archives to the current directory.
+     *  Decompress `tar.bz2` archives to [destination] directory.
      *
-     *  @return files processed
+     *  @param destination
      */
-    internal fun verboseUnzip(
+    internal fun unzip(
         connection: SshConnection,
         archive: String,
+        destination : String,
         timeout: Duration
-    ): Iterable<String> {
+    ) {
         ubuntu.install(connection, listOf("lbzip2"))
-        return time("unzip") {
-            connection.execute("tar -I lbzip2 -xvf $archive", timeout).output.splitToSequence("\n").asIterable()
+        time("unzip") {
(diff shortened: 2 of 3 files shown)


src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/dataset/FileArchiver.kt
```diff
@@ -23,18 +23,20 @@ class FileArchiver {
     }
 
     /**
-     *  Decompress `tar.bz2` archives to the current directory.
+     *  Decompress `tar.bz2` archives to `destination` directory.
```
dagguh: You can link it via `[destination]`

CHANGELOG.md
```diff
@@ -26,6 +26,10 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/infrastructure/compare/release-4.12.4...master
 
+### Fixed
+- Fix dataset unzipping performance. Fix [JPERF-504].
+
+[JPERF-504]: https://ecosystem.atlassian.net/browse/JPERF-504
 ## [4.12.4] - 2019-06-25
```
mcmurphy2000: Missing blank line

src/main/kotlin/com/atlassian/performance/tools/infrastructure/dataset/HttpDatasetPackage.kt
```diff
@@ -40,17 +40,32 @@ internal class HttpDatasetPackage(
...
     private fun unzip(ssh: SshConnection, resourceName: String, timeForUnzipping: Duration): String {
-        val topPaths = FileArchiver()
-            .verboseUnzip(ssh, resourceName, timeForUnzipping)
-            .asSequence()
-            .map { Paths.get(it) }
-            .map { it.first() }
-            .filter { it.toString().isNotBlank() }
+        val destination = UUID.randomUUID().toString()
+        ssh.execute("mkdir $destination")
+        FileArchiver().unzip(
```
dagguh: We're no longer consuming the returned `Iterable<String>`. Did the stdout streaming cause the slowdown?
If so, perhaps we should no longer return it from the API to prevent accidental use

src/main/kotlin/com/atlassian/performance/tools/infrastructure/dataset/HttpDatasetPackage.kt
```diff
@@ -40,17 +40,32 @@ internal class HttpDatasetPackage(
...
 
     private fun unzip(ssh: SshConnection, resourceName: String, timeForUnzipping: Duration): String {
-        val topPaths = FileArchiver()
-            .verboseUnzip(ssh, resourceName, timeForUnzipping)
-            .asSequence()
-            .map { Paths.get(it) }
-            .map { it.first() }
-            .filter { it.toString().isNotBlank() }
+        val destination = UUID.randomUUID().toString()
+        ssh.execute("mkdir $destination")
```
dagguh: So we're going for the Bogdan fix after all 😃

src/main/kotlin/com/atlassian/performance/tools/infrastructure/dataset/HttpDatasetPackage.kt
wyrzyk: Either stdout streaming or fetching it via ssh. It was never an API.

dagguh: Cool
WRT removing the `return`, it's internal, but still a ticking time bomb

dagguh:
So the problem was kinda caught by CI, we had to https://github.com/atlassian/infrastructure/commit/7fd506fbdfe2f8b36a8e13831ce3d2851892cc25
Shall we revert that?

PR #28, opened 2019-06-25
Issue/jperf 504 fix implementation

wyrzyk:
I was hit by a difference in our module's test environment (docker) vs AWS environment. We probably should address this problem to avoid mistakes like this in the future.

I'll run the test on a snapshot version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/infrastructure/compare/release-4.12.5...master
 
+### Fixed
+- Fix dataset unzipping permissions problem. Fix [JPERF-504].
+
 ## [4.12.5] - 2019-06-25
 [4.12.5]: https://github.com/atlassian/infrastructure/compare/release-4.12.4...release-4.12.5
 
--- a/src/main/kotlin/com/atlassian/performance/tools/infrastructure/Ls.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/infrastructure/Ls.kt
@@ -1,0 +1,17 @@
+package com.atlassian.performance.tools.infrastructure
+
+import com.atlassian.performance.tools.ssh.api.SshConnection
+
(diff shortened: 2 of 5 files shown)


wyrzyk:
I need to remove mark commit :)

dagguh:
This was not caught due to Docker env user being `root` and not having to `sudo` for anything.
We can cross fingers and hope this kind of a bug won't happen again. But if we do cover it, then not only do we prove that this fix works, but we'd gain immunity to `sudo` bugs for the future.

Testing with a snapshot is a convenient workaround, but it leaves the underlying issue unsolved. It also sets a precedent to always devolve to snapshot testing, further degrading release confidence and quality.

CHANGELOG.md
```diff
@@ -26,6 +26,9 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/infrastructure/compare/release-4.12.5...master
 
+### Fixed
+- Fix dataset unzipping permissions problem. Fix [JPERF-504].
```
dagguh: The original JPERF-504 symptom is different. It's present in 4.12.4 and fixed in 4.12.5
The new one is present in 4.12.5 and is going to be fixed in 4.12.6
> Error while executing mv df54f5f1-881a-4754-9c83-8adfc2ad9ff8/database ... Exit status code SshResult(exitStatus=1, output=, errorOutput=mv: cannot create directory '../database': Permission denied

JPERF-504 does contain the symptom in the comments, so it is findable, so it's okayish

dagguh: Shall we revert https://github.com/atlassian/infrastructure/pull/25 if JPERF-504 is indeed fixed?

wyrzyk: The change doesn't affect the test duration much, and we were very close to the limit before. I'd keep the new limit.

wyrzyk: https://ecosystem.atlassian.net/browse/JPERF-505

wyrzyk: ^^ refers to ```This was not caught due to Docker env ...```

PR #30, opened 2019-07-05
JPERF-519: Facilitate Jira license overrides

dagguh:


src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/LicenseOverridingDatabase.kt
```diff
@@ -0,0 +1,38 @@
...
+import com.atlassian.performance.tools.infrastructure.database.SshMysqlClient
+import com.atlassian.performance.tools.ssh.api.SshConnection
+import java.net.URI
+
+/**
+ * Removes all licenses from [database] and adds [licenses] instead.
+ *
+ * @since 4.13.0
+ */
+class LicenseOverridingDatabase(
```
wyrzyk: Isn't it `MySqlLicenseOverridingDatabase`?

dagguh: Good point

wyrzyk:
1. The test doesn't work on CircleCi.
2. Tests started to run on one fork :disappointed: and build time increased from 12 minutes to > 1h.

PR #32, opened 2019-07-11
QUICK-224 deprecate the use of strings to hold and transfer jira licenses

mminns:
Pulls in code that was first implemented as workarounds/improvements in https://github.com/atlassian/jira-hardware-exploration


mminns:
Oops I realize I referenced the wrong QUICK in the commit messages etc should be 221 **not** 224 I will correct that

dagguh:
Raised https://ecosystem.atlassian.net/browse/JPERF-523

PR #34, opened 2019-07-22
JPERF-531: Add support for wall-clock profiler

wyrzyk:
I'm using the implementation internally, but have no integration tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ### Fixed
 - Increase timeout of apt-get. Fix [JPERF-552]
 
+### Added
+- Add support for wall-clock profiler. Resolves [JPERF-531].
+
+[JPERF-531]: https://ecosystem.atlassian.net/browse/JPERF-531
 [JPERF-552]: https://ecosystem.atlassian.net/browse/JPERF-552
 
 ## [4.14.1] - 2019-07-24
--- a/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/profiler/WallClockProfiler.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/profiler/WallClockProfiler.kt
@@ -1,0 +1,37 @@
+package com.atlassian.performance.tools.infrastructure.api.profiler
+
+import com.atlassian.performance.tools.infrastructure.api.process.RemoteMonitoringProcess
+import com.atlassian.performance.tools.ssh.api.SshConnection
+
+/**
+ * A profiler to sample all threads equally every given period of time regardless of thread status.
+ */
+class WallClockProfiler : Profiler {
+    override fun install(ssh: SshConnection) {
+        ssh.execute("wget -q https://github.com/jvm-profiling-tools/async-profiler/releases/download/v1.5/async-profiler-1.5-linux-x64.tar.gz")
+        ssh.execute("mkdir async-profiler")
+        ssh.execute("tar -xzf async-profiler-1.5-linux-x64.tar.gz -C async-profiler")
+        ssh.execute("sudo sh -c 'echo 1 > /proc/sys/kernel/perf_event_paranoid'")
+        ssh.execute("sudo sh -c 'echo 0 > /proc/sys/kernel/kptr_restrict'")
+    }
+
+    override fun start(
+        ssh: SshConnection,
+        pid: Int
+    ): RemoteMonitoringProcess {
+        ssh.execute("./async-profiler/profiler.sh -e wall -i 9ms -b 160000000 start $pid")
+        return ProfilerProcess(pid)
+    }
+
+    private class ProfilerProcess(private val pid: Int) : RemoteMonitoringProcess {
+        private val flameGraphFile = "wall-clock-flamegraph.svg"
+
+        override fun stop(ssh: SshConnection) {
+            ssh.execute("./async-profiler/profiler.sh stop $pid -o svg > $flameGraphFile")
+        }
+
+        override fun getResultPath(): String {
+            return flameGraphFile
+        }
+    }
+}


src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/profiler/WallClockProfiler.kt
```diff
@@ -0,0 +1,37 @@
...
+    ): RemoteMonitoringProcess {
+        ssh.execute("./async-profiler/profiler.sh -e wall -i 9ms -f -b 160000000 start $pid")
+        return ProfilerProcess(pid)
+    }
+
+    private class ProfilerProcess(private val pid: Int) : RemoteMonitoringProcess {
+        private val flameGraphFile = "flamegraph.svg"
+
+        override fun stop(ssh: SshConnection) {
+            ssh.execute("./async-profiler/profiler.sh stop $pid -o svg > $flameGraphFile")
```
dagguh: Can we see the thread states in the flamegraph? If not, is there any way?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/profiler/WallClockProfiler.kt
```diff
@@ -0,0 +1,37 @@
...
+    override fun start(
+        ssh: SshConnection,
+        pid: Int
+    ): RemoteMonitoringProcess {
+        ssh.execute("./async-profiler/profiler.sh -e wall -i 9ms -f -b 160000000 start $pid")
+        return ProfilerProcess(pid)
+    }
+
+    private class ProfilerProcess(private val pid: Int) : RemoteMonitoringProcess {
+        private val flameGraphFile = "flamegraph.svg"
```
dagguh: Is it legit to combine `AsyncProfiler` with `WallClockProfiler` at the same time?

CHANGELOG.md
```diff
@@ -26,6 +26,9 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/infrastructure/compare/release-4.13.0...master
```
dagguh: FYI [4.14.0](https://packages.atlassian.com/maven-public/com/atlassian/performance/tools/infrastructure/4.14.0/) was [released](https://travis-ci.com/atlassian/infrastructure/builds/119456491) on 2019-07-17.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/profiler/WallClockProfiler.kt
wyrzyk: I'm not sure. I would not suggest doing that.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/profiler/WallClockProfiler.kt
wyrzyk: Nope, this way we gather all the threads. But you can use async-profiler for different `events` to track what you want (running threads/locks/context-switches).

CHANGELOG.md
```diff
@@ -34,6 +37,7 @@ Dropping a requirement of a major version of a dependency is a new contract.
 
 
 [JPERF-523]: https://ecosystem.atlassian.net/browse/JPERF-523
+[JJPERF-531]: https://ecosystem.atlassian.net/browse/JPERF-531
```
mcmurphy2000: `[JJPERF-531]` -> `[JPERF-531]`

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/profiler/WallClockProfiler.kt
```diff
@@ -0,0 +1,37 @@
...
+        ssh.execute("tar -xzf async-profiler-1.5-linux-x64.tar.gz -C async-profiler")
+        ssh.execute("sudo sh -c 'echo 1 > /proc/sys/kernel/perf_event_paranoid'")
+        ssh.execute("sudo sh -c 'echo 0 > /proc/sys/kernel/kptr_restrict'")
+    }
+
+    override fun start(
+        ssh: SshConnection,
+        pid: Int
+    ): RemoteMonitoringProcess {
+        ssh.execute("./async-profiler/profiler.sh -e wall -i 9ms -f -b 160000000 start $pid")
```
wyrzyk: I may need to adjust the buffer.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/profiler/WallClockProfiler.kt
wyrzyk: Can you explain the use case? I don't feel I have a full context to answer your question. 

Please check an example flame graph I've attached to the public ticket. This is the output you should expect. 

It'll aggregate state per multiple threads. You can use `async-profiler` with `-t` option to profile threads separately.

dagguh: The use case is bottleneck analysis.
Let's say 30% of actions are slow and bring Apdex down significantly. I can filter action metrics to narrow them down to the slow ones, then I can analyze their waterfalls and let's say most of them point to slow server. Then I need to profile server, but not it's entirety, but only the periods of time, when the slow actions were run. I don't want to include the stacktraces and thread states of the time where server was only serving quick actions.

wyrzyk: @dagguh do you mean something like heat maps?
https://medium.com/netflix-techblog/netflix-flamescope-a57ca19d47bb 
 http://www.brendangregg.com/HeatMaps/subsecondoffset.html

dagguh: Yes! This is brilliant!

dagguh:
@wyrzyk should I wait for this or can I release `master` right now?

dagguh:
I'm gonna go ahead and release

dagguh:
GH features came in handy. The force-push diff showed me what the rebase did and the "files viewed" filter showed me that only the CHANGELOG has changed since last I saw.

PR #38, opened 2019-08-20
JPERF-552 : Increase apt-get update timeout

lipinskipawel:


src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/browser/chromium/Chromium69IT.kt
```diff
@@ -17,7 +21,11 @@ class Chromium69IT {
 
                 Chromium69().install(connection)
 
-                val installedAfter = isChromiumInstalled(connection)
+                var installedAfter = false
```
lipinskipawel: All those tricks are workarounds. Right now we have a race between starting some process and the client who wants to start using this. We should improve `DetachedProcess` class e.g. the `start` method should be blocking one or returns CompletableFuture. Either way I think is fair to create a JPERF ticket for that.

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/ThreadDumpIT.kt
```diff
@@ -19,7 +22,10 @@ class ThreadDumpIT {
                 sshConnection.execute("${jdk.use()}; javac Test.java")
                 val process = sshConnection.startProcess("${jdk.use()}; java Test")
                 try {
-                    val pid = getPid(sshConnection, jdk)
+                    var pid = -1
```
dagguh: Same here, no need for temporary mutable values

src/test/kotlin/com/atlassian/performance/tools/infrastructure/browser/SshChromium.kt
```diff
@@ -1,14 +1,15 @@
...
 
+import com.atlassian.performance.tools.infrastructure.api.jvm.StaticBackoff
+import com.atlassian.performance.tools.jvmtasks.api.IdempotentAction
 import com.atlassian.performance.tools.ssh.api.DetachedProcess
 import com.atlassian.performance.tools.ssh.api.SshConnection
 import com.atlassian.performance.tools.virtualusers.api.browsers.Browser
 import com.atlassian.performance.tools.virtualusers.api.browsers.CloseableRemoteWebDriver
 import org.openqa.selenium.chrome.ChromeOptions
 import org.openqa.selenium.remote.RemoteWebDriver
-import org.rnorth.ducttape.unreliables.Unreliables.retryUntilSuccess
```
dagguh: Woo! Can we remove the dependency from Gradle?

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/browser/chromium/Chromium69IT.kt
dagguh: Hell yeah, fighting the flakiness! 💪

src/test/kotlin/com/atlassian/performance/tools/infrastructure/browser/SshChromium.kt
```diff
@@ -27,14 +28,23 @@ internal class SshChromium(
...
     }
-}
\ No newline at end of file
+
+    private fun waitForChromeProcess(ssh: SshConnection) {
+        ssh.safeExecute("ps aux")
+            .output
+            .split("\n")
+            .filter { it.contains("./chromedriver --whitelisted-ips") }
+            .single { it.length < 100 }
```
wyrzyk: Why do you filter by length? Why 100?

lipinskipawel: Because there are two processes'es with the `./chromedriver --whitelisted-ips` command and we are looking for the second process, which has fewer characters.

wyrzyk: Is it a reliable assumption?

lipinskipawel: I suspect that this is more reliable than before. But I have no idea how to measure this right now.

PR #42, opened 2019-09-16
Issue/jperf 571 fix infrastructure ci

wyrzyk:


build.gradle.kts
```diff
@@ -85,6 +95,10 @@ tasks.getByName("test", Test::class).apply {
 }
 
 val testIntegration = task<Test>("testIntegration") {
+    testLogging {
+        events = setOf(TestLogEvent.PASSED, TestLogEvent.SKIPPED, TestLogEvent.FAILED)
```
dagguh: It's no longer needed with scans enabled.

src/test/kotlin/com/atlassian/performance/tools/infrastructure/SshUbuntuExtensions.kt
```diff
@@ -15,7 +15,7 @@ internal fun SshUbuntu.toSsh(): Ssh {
     })
     ssh.newConnection().use { connection ->
         connection.execute("apt-get update -qq", Duration.ofMinutes(3))
-        connection.execute("apt-get install sudo curl screen gnupg2 -y -qq")
+        connection.execute("export DEBIAN_FRONTEND=noninteractive; apt-get install sudo curl screen gnupg2 -y -qq", Duration.ofMinutes(5))
```
dagguh: We could reuse `Ubuntu` here

.travis.yml
```diff
@@ -6,7 +6,7 @@ stages:
 jobs:
   include:
     - stage: test
-      script: travis_wait 25 ./gradlew build
+      script: travis_wait 45 ./gradlew build
```
dagguh: Green 45 minutes is way better than red 25 minutes. It's worth the tradeoff.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/browser/Chrome.kt
```diff
@@ -16,7 +16,7 @@ class Chrome : Browser {
     override fun install(ssh: SshConnection) {
         ssh.execute("wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | sudo apt-key add")
         ssh.execute("echo 'deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main' | sudo tee -a /etc/apt/sources.list.d/google-chrome.list")
-        Ubuntu().install(ssh, listOf("google-chrome-stable"), ofMinutes(3))
+        Ubuntu().install(ssh, listOf("google-chrome-stable"), ofMinutes(5))
```
dagguh: Bumping the timeouts all over is a good move. These should work for poor internet bandwidth as well. Nice touch with including the rationale in the commit message: bfa39ba5116c872179e91412fffec3cc57797418

src/test/kotlin/com/atlassian/performance/tools/infrastructure/SshUbuntuExtensions.kt
wyrzyk: We can't. Ubuntu assumes `sudo` is already installed.

.travis.yml
wyrzyk: It's around 35 minutes right now, but I run it only once and I don't have better data to set it correctly.

wyrzyk: > Successful in 26m. 

It would be great to gather some statistics and decrease the timeout.

dagguh: Gradle Enterprise might have a feature for that

PR #44, opened 2019-10-14
JPERF-219: Fix apt-get flakes

wyrzyk:


dagguh:
The fix flaked: https://scans.gradle.com/s/emnaqui57uujs/tests/2tfdhyyojg4t4-lelh3rsl4mr2w
```
2019-10-14T13:32:15,916Z DEBUG Test worker [] [com.atlassian.performance.tools.ssh.SshjConnection] root$ kill $(pidof apt-get)
2019-10-14T13:32:15,925Z DEBUG Test worker [] [com.atlassian.performance.tools.ssh.SshjConnection] kill: usage: kill [-s sigspec | -n signum | -sigspec] pid | jobspec ... or kill -l [sigspec]
...
... SshResult(exitStatus=100, output=, errorOutput=E: dpkg was interrupted, you must manually run 'sudo dpkg --configure -a' to correct the problem.
```
Firstly, the "kill usage" is a WTF factor, we should do explicit `pidof` presence check.
Secondly, we might need to nuke the locks after all. Perhaps only when the process is already absent.

dagguh:
Already shipped via #45 and #49

PR #45, opened 2019-10-18
JPERF-587: Refactor integration tests

wyrzyk:
The tests take ~35 minutes now. It's too much for a comfortable dev-loop and it's also close to the hard limit of 50 minutes on Travis CI. I aim to reduce the time by removing redundant or not critical tests.


src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/storage/PublicJiraSoftwareDistributionsTest.kt
```diff
@@ -0,0 +1,24 @@
...
+
+import com.atlassian.performance.tools.infrastructure.api.distribution.PublicJiraSoftwareDistribution
+import com.atlassian.performance.tools.infrastructure.mock.RememberingSshConnection
+import org.assertj.core.api.Assertions
+import org.junit.Test
+
+class PublicJiraSoftwareDistributionsTest {
+
+    @Test
+    fun shouldDelegateToPublicJiraSoftwareDistribution() {
```
dagguh: Woop woop!
I love it when we can replace an integration test with a unit one without losing coverage 👍

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/storage/PublicJiraSoftwareDistributionsTest.kt
```diff
@@ -0,0 +1,24 @@
...
+        @Suppress("DEPRECATION") val jiraDistributionFromOldApi: ProductDistribution = PublicJiraSoftwareDistributions().get(jiraSoftwareVersion)
+        val jiraDistribution = PublicJiraSoftwareDistribution(jiraSoftwareVersion)
+
+        val sshOldApi = RememberingSshConnection()
+        jiraDistributionFromOldApi.install(sshOldApi, "destination")
+
+        val sshNewApi = RememberingSshConnection()
+        jiraDistribution.install(sshNewApi, "destination")
+
+        Assertions.assertThat(sshOldApi.commands).containsAll(sshNewApi.commands)
```
dagguh: ❤️

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/dataset/HttpDatasetPackageIT.kt
```diff
@@ -17,8 +17,8 @@ class HttpDatasetPackageIT {
     @Test
     fun shouldDownloadDataset() {
         val dataset = HttpDatasetPackage(
-            uri = URI("https://s3-eu-west-1.amazonaws.com/jpt-custom-datasets-storage-a008820-datasetbucket-1sjxdtrv5hdhj/af4c7d3b-925c-464c-ab13-79f615158316/database.tar.bz2"),
-            downloadTimeout = Duration.ofMinutes(2)
+            uri = URI("https://jpt-custom-datasets-storage-a008820-datasetbucket-1sjxdtrv5hdhj.s3-eu-west-1.amazonaws.com/mock-database.tar.bz2"),
```
dagguh: Nice and easy

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/browser/chromium/Chromium69.kt
```diff
@@ -14,7 +15,9 @@ class Chromium69 : Browser {
      * Installs chromium 69 with a compatible chromedriver.
      */
     override fun install(ssh: SshConnection) {
-        ChromiumInstaller(chromiumUri).install(ssh)
-        ChromedriverInstaller(chromedriverUri).install(ssh)
+        ParallelExecutor().execute(
```
dagguh: Nice.
`Ubuntu.install` does mutex internally, but other operations can definitely benefit from parallelization.

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/os/UbuntuIT.kt
```diff
@@ -27,6 +31,54 @@ class UbuntuIT {
         executor.shutdownNow()
     }
 
+    @Test
+    fun shouldRetry() {
```
dagguh: Hope it works 🤞 
After my refactor I saw Gradle exit code 137 AKA kernel-level OOME

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/browser/chromium/Chromium69.kt
wyrzyk: I'd go for the second option as it's easier. We can cover the packages installation with a unit test.

dagguh: WRT 2: we'd lose coverage, increase complexity and optimize the test only, that's a no for me

WRT 1: we can use the dynamic apt-get mirrors

dagguh: > We can cover the packages installation with a unit test.
But the package installation installs some OS-level libs, which are required by the subsequent Chrome installation. I remember that without the dependencies apt-get fails and informs us about which libs are missing. If the chrome changes dependencies, the unit test won't show it.

wyrzyk: IT would still catch the problem you mention, but you're probably right it's not worth the extra complexity. I've just checked and the extra dependencies weight ~354 MB.

```
atlassian/ssh-ubuntu     0.2.3-SNAPSHOT      1db502ad858f        31 minutes ago      609MB
atlassian/ssh-ubuntu     0.2.2               7765843907b0        4 weeks ago         255MB
```

gradle.properties
```diff
@@ -0,0 +1,2 @@
+org.gradle.parallel=true
+org.gradle.workers.max=4
```
dagguh: 18 minutes loop is quite good

wyrzyk: https://scans.gradle.com/s/za7v4kgvzwqjw

dagguh: 9 minutes! Sweet

PR #47, opened 2019-11-21
random acts of hardening

pbruski:


src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/os/Ubuntu.kt
```diff
@@ -29,7 +29,7 @@ class Ubuntu {
             )
         }
             .retry(
-                maxAttempts = 2,
+                maxAttempts = 7, //we need to accommodate cron-based image updates happening in the background
```
dagguh: This hurts very with the `ExponentialBackoff` in place:
```
Attempt #5 failed to install [docker-ce=17.09.0~ce-0~ubuntu], Failed an attempt to install [docker-ce=17.09.0~ce-0~ubuntu], backing off for PT2M40S ...
```
Raised https://ecosystem.atlassian.net/browse/JPERF-619

PR #51, opened 2019-12-04
improved logging messages

pbruski:


src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/virtualusers/MulticastVirtualUsers.kt
```diff
@@ -52,16 +57,29 @@ class MulticastVirtualUsers<out T : VirtualUsers>(
             throw Exception("$virtualUsers virtual users are not enough to spread into $nodeCount nodes")
         }
         val loadSlices = load.slice(nodeCount)
+        val activeNodes = ConcurrentHashMap.newKeySet<String>()
+
+        val roughTotalTime = options.behavior.load.total.plusSeconds(59)
```
wyrzyk: Why + 59?

dagguh:
```suggestion
        val roughTotalTime = load.total + behavior.maxOverhead
```

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/virtualusers/MulticastVirtualUsers.kt
```diff
@@ -52,16 +57,29 @@ class MulticastVirtualUsers<out T : VirtualUsers>(
             throw Exception("$virtualUsers virtual users are not enough to spread into $nodeCount nodes")
         }
         val loadSlices = load.slice(nodeCount)
+        val activeNodes = ConcurrentHashMap.newKeySet<String>()
```
dagguh:
```suggestion
        val activeNodes = ConcurrentHashMap.newKeySet<T>()
```
And then we don't have `toString` later on. We can rely on `equals` and `hashCode` rather than hoping for unique `toString` values. All known `VirtualUsers` impls have identity equality. And I don't foresee a case, where different physical VU nodes should be considered equal, so no future impl should hurt us here.

dagguh:
Huh, the CI failed with:
```
Error while executing apt-get update -qq. Exit status code SshResult(exitStatus=100, output=, errorOutput=E: The repository 'mirror://mirrors.ubuntu.com/mirrors.txt bionic Release' no longer has a Release file.
```
The error message suggests something external has changed with time. [Spinning master](https://travis-ci.com/atlassian/infrastructure/builds/139492288) to check that.

dagguh:
It was a flake, yuck
So I reran the PR CI and it also went green

dagguh:
I'll merge this as is, because it's valuable
I'll do the refactor myself in another PR

PR #54, opened 2019-12-13
add timeouts and retries to lftp

pbruski:


src/main/kotlin/com/atlassian/performance/tools/infrastructure/HttpResource.kt
```diff
@@ -27,7 +27,7 @@ class HttpResource(
     ) {
         Ubuntu().install(ssh, listOf("lftp"), Duration.ofMinutes(2))
         ssh.execute(
-            """lftp -c 'pget -n 64 -c "$uri" -o $destination'""",
+            """lftp -c 'set net:timeout 15; set net:max-retries 10; pget -n 32 -c "$uri" -o $destination'""",
```
dagguh: [`man lftp`](https://linux.die.net/man/1/lftp) :
> net:timeout (time interval)
    sets the network protocol timeout.

Is it in seconds?

dagguh: AFAIK this aims to fix SSH timeouts in `Chromium69`. Could you leave a mark in the changelog?

PR #56, opened 2020-02-17
JPERF-466: Use non-deprecated Kotlin stdlib

wyrzyk:


wyrzyk:
@mzyromski-atlassian Github doesn't recognize you as a code owner. Here's PR to solve the issue: https://github.com/atlassian/infrastructure/pull/57

PR #60, opened 2020-03-03
NONE: Add Browser's KDocs

wyrzyk:


src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/browser/Browser.kt
```diff
@@ -2,6 +2,15 @@ package com.atlassian.performance.tools.infrastructure.api.browser
 
 import com.atlassian.performance.tools.ssh.api.SshConnection
 
+/**
+ * Installs browser and WebDriver. See [Chrome] or [com.atlassian.performance.tools.infrastructure.api.browser.chromium.Chromium69].
+ * Browser must provide public no-args constructor.
```
dagguh: They don't have to provide public constructor.
We only require this no-arg stuff, where we're forced to use reflection, that is in a cross-JVM communication between the local JVM and the remote VU JVMs. So only the VU SPIs require the nullary constructor.

CHANGELOG.md
```diff
@@ -26,6 +26,9 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/infrastructure/compare/release-4.14.5...master
 
+### Added
+- `com.atlassian.performance.tools.infrastructure.api.browser.Browser` KDocs.
```
dagguh: Technically it adds behavioral contracts:
* `Browser` has to install WebDriver (we don't need this contract IMHO)
* `SshConnection` targets Ubuntu - this one is critical for implementors, so we need it

mzyromski-atlassian: @dagguh Are you suggesting this should be explicitly stated in the changelog?

dagguh: Yeah, that's the diff.
I wanna empathize with lib consumers. As a consumer of a lib, I'm gonna look for useful stuff, e.g. "since when can I rely on it being Ubuntu?". Not gonna look for "when did KDocs appear"?

Docs themselves are a standalone description and changelog is a series of human-level diffs.

PR #62, opened 2020-03-27
Test on CircleCI for speed

dagguh:
Gain more cross-module consistency as a bonus.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,53 @@
+version: 2.1
+
+jobs:
+  build:
+    machine: true
+    environment:
+      TERM: dumb
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - v1-dependencies-{{ checksum "build.gradle.kts" }}
+            - v1-dependencies-
+      - run: ./gradlew compileTestKotlin
+      - save_cache:
+          paths:
+            - ~/.gradle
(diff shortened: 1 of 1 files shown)


dagguh:
Example: https://app.circleci.com/pipelines/github/dagguh/infrastructure/3/workflows/1d9e3dba-5353-4604-b063-b76814f97892/jobs/8

mzyromski-atlassian:
The builds on CircleCI seem actually to take longer than on Travis (~10 mins). Do you know why that's the case? 

On the other hand consistency is a compelling argument.

dagguh:
3ef0383 is from December 2019, [back then](https://travis-ci.com/github/atlassian/infrastructure/builds) Travis took ~20 minutes to run.
image

They must have changed something in their infra. The only diff in our repo is 3405bccb2bf9032fa071d40580454c1fe6c3ab4b and 509a031928f2f60f16d15487b66329bd9bcf1734, which don't explain the build time diff.

Let's merge it for consistency and artifacts?

mzyromski-atlassian:
Sure, let's give it a go. I guess we should then also remove Travis config file, shouldn't we?

dagguh:
We can remove Travis only after merging this.
Otherwise, this PR won't pass the existing CI checks.

PS. Please add a review. I cannot merge without an approve. If I need to do something more, you can "request changes". If we can merge as is, "approve". If there are some things that are nice to have, but we can merge without them, add task lists and approve.

PR #63, opened 2020-03-30
Start Docker after it's installed

dagguh:


mzyromski-atlassian:
Looks like the newly introduced `DockerIT` test takes a while and moves the entire test suite execution time over 20 minutes which is a timeout defined on CI.

dagguh:
Left to do:
- [ ] rebase onto master to get CircleCI
- [ ] fix https://ecosystem.atlassian.net/browse/JPERF-619 by consuming https://github.com/atlassian/jvm-tasks/pull/2 and https://github.com/atlassian/jvm-tasks/pull/3 - new 🔴 commit
- [ ]  fix https://ecosystem.atlassian.net/browse/JPERF-425 by parametrizing `Docker` with it's apt-get version- new commit - new 💚 commit
- [ ] let consumers use the fix: expose `DockerImage` as API and parametrize it with `Docker`, which should also become API - new 💚 commit
- [ ] cascade the parametrization to `DockerImage` consumers : `MySqlDatabase`, `AtlassianSplunkForwarder`, `UniversalSplunkForwarder` - new 💚 commit

Explicit parametrization requires the cascade and the consumers to match `Docker.version` with their Ubuntu image. It gives a lot of control at the cost of inconvenience.
If we want to maximize convenience at the expense of control, we can avoid exposing any new APIs if we're able to select a proper `"docker-ce=$version"` in `Docker` impl.

dagguh:
[`DockerIT` was introduced](https://github.com/atlassian/infrastructure/commit/24b0c5882c73b49d8cb174d8d42cfa86bde20bd8#diff-6bf16fa7261e87b8204856025b39cfb6a9b81adb7a7497321281bb2fd2940b2f) in the meantime and CI is green

PR #64, opened 2020-05-11
JPERF-625 fix Docker compatibility with Bionic+ and upgrade Docker to 19.03

pbruski:


dagguh:
Welp, it's been a bit rushed IMHO. No discussion about compatibility and SemVer

src/main/kotlin/com/atlassian/performance/tools/infrastructure/Docker.kt
```diff
@@ -24,15 +24,16 @@ internal class Docker {
             ),
             timeout = Duration.ofMinutes(2)
         )
+        ubuntu.addKey(ssh, "7EA0A9C3F273FCD8")
+
         val release = ssh.execute("lsb_release -cs").output
-        val repository = "deb [arch=amd64] https://download.docker.com/linux/ubuntu $release stable"
-        ssh.execute("sudo add-apt-repository \"$repository\"")
-        ssh.execute("curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -")
-        val version = "17.09.0~ce-0~ubuntu"
```
dagguh, line 31: So it will stop working with Ubuntu 17. Consumers will have to switch to Ubuntu 19, including `aws-infrastructure`, `virtual-users` CI, and an unknown number of other use cases - maybe HWRs?
We convert this this de-facto dependence into a behavioral contract and then mark this as breaking.

dagguh: Or we could try to work with a range of Ubuntu versions.

pbruski: It's not an Ubuntu version, it's a Docker version. These changes have been tested with Xenial. That Docker version is available on Xenial.

dagguh: Nice! It improves the situation a lot. Then it it indeed a transparent change - `Fixed`/patch.

pbruski:
> Welp, it's been a bit rushed IMHO. No discussion about compatibility and SemVer

It will only break if Docker change breaks. So this is a choice between a forced semver change (because beyond Disco, the only supported version is 19) or a semver change if we get a bug report (which will likely never happen, we use pretty rudimentary options of Docker).

src/main/kotlin/com/atlassian/performance/tools/infrastructure/Docker.kt
mzyromski-atlassian: See #65

mzyromski-atlassian:
Released in `4.15.1`.

PR #70, opened 2020-06-02
Mark 4.16.0 release

mzyromski-atlassian:

PR #71, opened 2020-06-02
Fix codename detection

pbruski:


src/main/kotlin/com/atlassian/performance/tools/infrastructure/Docker.kt
```diff
@@ -19,14 +19,13 @@ internal class Docker {
             packages = listOf(
                 "apt-transport-https",
                 "ca-certificates",
-                "curl",
-                "software-properties-common"
```
pbruski, line 23: this is 150MBs to get one line for a text file...

src/test/kotlin/com/atlassian/performance/tools/infrastructure/DockerIT.kt
```diff
@@ -0,0 +1,22 @@
...
+import org.junit.Test
+import java.time.Duration
+
+
+class DockerIT {
+    @Test
+    fun installWorks() {
+        SshUbuntuContainer().start().use { ssh ->
+            ssh.toSsh().newConnection().use { connection ->
+                //workaround for a bug in Docker download site for bionic
```
pbruski, line 13: This sucks. I think we should have a helper class that sets up this container from a pristine ubuntu container. I'll spike it.

mzyromski-atlassian:
@pbruski, could you please add a CHANGELOG entry?

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ Dropping a requirement of a major version of a dependency is a new contract.
 
 ### Fixed
 - Eliminate spurious apt warnings. Fix [JPERF-642](https://ecosystem.atlassian.net/browse/JPERF-642).
+- Corrected Ubuntu codename detection
```
wyrzyk, line 35: Can you also take care of the commit message?

mzyromski-atlassian: Also, sorry for that, but in the meantime #70 was merged and this entry should go into Unreleased section. Could you please rebase?

mzyromski-atlassian:
Released in `4.16.1`.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/Docker.kt
dagguh: Wow! Great catch

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/os/Ubuntu.kt
```diff
@@ -100,4 +100,8 @@ class Ubuntu {
         ssh.execute("echo '$repository' | sudo tee /etc/apt/sources.list.d/${sourceFileName}.list")
         updatePackageIndex(ssh)
     }
+
+    fun getDistributionCodename(ssh: SshConnection): String {
```
dagguh, line 104: New API FYI @mzyromski-atlassian

dagguh:
> Released in `4.16.1`.

Should be `4.17.0`, new API was added.

mzyromski-atlassian:
> Should be `4.17.0`, new API was added.

True, missed that. :( Would you advise releasing `4.17.0` now?

dagguh:
Yes, it will allow the new API consumers to depend on 4.17.0+ and still being able to downgrade patch versions in peace.

PR #73, opened 2020-06-03
Mark 4.17.0 release

mzyromski-atlassian:
Artifacts: https://packages.atlassian.com/maven-public/com/atlassian/performance/tools/infrastructure/4.17.0/

4.16.1 shouldn't have been released as recent changes demand a minor release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,13 +24,21 @@ Adding a requirement of a major version of a dependency is breaking a contract.
 Dropping a requirement of a major version of a dependency is a new contract.
 
 ## [Unreleased]
-[Unreleased]: https://github.com/atlassian/infrastructure/compare/release-4.16.0...master
+[Unreleased]: https://github.com/atlassian/infrastructure/compare/release-4.17.0...master
+
+## [4.17.0] - 2020-06-03
+[4.17.0]: https://github.com/atlassian/infrastructure/compare/release-4.16.0...release-4.17.0
+
+### Fixed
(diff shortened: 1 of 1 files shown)


mzyromski-atlassian:
@dagguh, WDYT: should we keep `4.16.1` entry in the changelog?

dagguh:
We should keep it. Hiding releases won't work, because it already spread via Maven repos, mirrors and local repos/caches. People will stumble upon that release. It's helpful to describe what it brings.

The nuance is: new API in 4.16.1 cannot be relied on (e.g. we can release 4.16.2 without it), but it can be relied on since 4.17.0.

CHANGELOG.md
```diff
@@ -24,14 +24,17 @@ Adding a requirement of a major version of a dependency is breaking a contract.
 Dropping a requirement of a major version of a dependency is a new contract.
 
 ## [Unreleased]
-[Unreleased]: https://github.com/atlassian/infrastructure/compare/release-4.16.0...master
+[Unreleased]: https://github.com/atlassian/infrastructure/compare/release-4.17.0...master
 
-## [4.16.1] - 2020-06-03
-[4.16.1]: https://github.com/atlassian/infrastructure/compare/release-4.16.0...release-4.16.1
```
dagguh, line 30: Let's preserve the 4.16.1 entry.
We made such mistake before: https://github.com/atlassian/report/commit/66d028197d9414587adc8eacaff669ffbb90132e

PR #75, opened 2020-07-27
JPERF-449: Expect Docker images to be runnable in CI

dagguh:

PR #76, opened 2020-08-06
Add CircleCI status badge

dagguh:
The real reason is to trigger a build on the atlassian fork.


mzyromski-atlassian:
Are we switching to CircleCI now? If so:
* what's the rationale?
* shouldn't we first reconfigure merge checks?

dagguh:
I want to switch away from TravisCI, because it fails with privileged Docker containers: https://github.com/atlassian/infrastructure/pull/75#issuecomment-669859641

PR #79, opened 2020-08-06
Don't fork for every test

dagguh:
"One" is the least optimal value.
https://docs.gradle.org/current/userguide/java_testing.html :
> Warning: a low value (other than 0) can severely hurt the performance of the tests

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -102,7 +102,6 @@ val testIntegration = task<Test>("testIntegration") {
     filter {
         include("**/*IT.class")
     }
-    setForkEvery(1)
     maxParallelForks = 5
 }
 


dagguh:
The current bottleneck on GHA is installing Chromium dependencies. It takes 15 minutes, times out, and the second attempt works (after 9 minutes!). In CircleCI it takes 2 minutes. In Travis it also must be fast, because the total build time is 7 minutes - 2 minutes faster than Circle.

wyrzyk:
@dagguh Do you know why it was set to 1?

wyrzyk (dismissed):

dagguh:
> @dagguh Do you know why it was set to 1?

Git says cd6dece8f6f0dc73fdc5d82f5ca5f97f023a5c6b:
> Use 5 parallel forks

Perhaps it was used based on an interpretation of its name.

dagguh:
Superseded by #80

PR #81, opened 2020-08-07
JPERF-449: Docker on multiple Ubuntu versions

dagguh:


src/test/kotlin/com/atlassian/performance/tools/infrastructure/DockerIT.kt
```diff
@@ -1,31 +1,47 @@
...
 
-                Docker().install(connection)
-                DockerImage("hello-world").run(connection)
+                        Docker().install(ssh)
+                        DockerImage("hello-world").run(ssh)
+                    }
+                }
             }
-        }
+            .forEach { it.get() }
```
wyrzyk, line 32: No timeout? No need to shut down the pool?

src/test/kotlin/com/atlassian/performance/tools/infrastructure/sshubuntu/SshUbuntuImage.kt
```diff
@@ -0,0 +1,127 @@
...
+import java.time.Duration
+import java.util.UUID
+
+/**
+ * Supports:
+ * * [Ssh] to an Ubuntu with `sudo`
+ * * Docker "next to" Docker - able to start Docker instances even if the host is not privileged
+ * * granular Docker resource allocation and deallocation
+ */
+class SshUbuntuImage(
```
wyrzyk, line 27: Any plans to extract it for reuse?

dagguh: Yes, I want it to prove itself here, then extract it and import in `infrastructure` and `virtual-users`

src/test/kotlin/com/atlassian/performance/tools/infrastructure/DockerIT.kt
dagguh: 👍 fixed

dagguh:
Huh, I expected e75162e to be green. Locally I was testing `DockerIT` only, but it turns out other tests are red now. Good thing we got CI 😉

dagguh:
In the process of fixing other tests I found that the new SSH Ubuntu code had to install `screen` and `gnugpg` in the very first commit. After doing that, turns out that `DockerIT` just works, without any changes in `src/main` and regardless of Ubuntu version. It seems that JPERF-449 is already solved. The only value this PR now brings is:
- testing on both Ubuntu 18.04 and 16.04 (rastasheep doesn't start for 19, 20, dunno why)
- aligning the Docker install instructions with the official guide
- removing the `DockerIT` workaround

I'll reshape the changes to fit these points and push it anyway. I'll have to go back to unignoring `EntryPointIT` in `virtual-users` to rediscover what is the blocker now.

mgrzaslewicz:
I'd give it a try with our `ssh-ubuntu` library. It can be easily adjusted to have ubuntu version provided

mgrzaslewicz:
Some commit are lacking JPERF-449 prefix, should be reworded

dagguh:
> I'd give it a try with our ssh-ubuntu library. 

Ok, I'll try

mgrzaslewicz:
> Huh, I expected [e75162e](https://github.com/atlassian/infrastructure/commit/e75162e665c9057f0d9c819084f44cbc3d247dd9) to be green. Locally I was testing `DockerIT` only, but it turns out other tests are red now. Good thing we got CI wink
> 
>     * [x]  fix other tests

Is that a matter of not running tests locally or a difference between CI and local env? If the latter, could that be unified?

dagguh:
> Some commits are lacking JPERF-449 prefix, should be reworded

Some commits only fix tests. Tests are not released to consumers. So technically they're do not fix JPERF-449. Only `src/main` (or generated POM) changes do. These commits were needed to ship the fix, but they're not the fix itself.

What are the issue keys for? They're auto-linked in Jira. Are they for consumers or maintainers? Consumers already got a `Fix version` field and changelogs. Maintainers can find all the related commits via Git history (finding this pull request), but I maybe it's sometimes useful to read directly from Jira? Or maybe they're never used?

@mgrzaslewicz, do you have a use case in mind? I've got a vague feeling that your expectation has merit, so we can make it the convention for @atlassian/jpt :+1:

dagguh:
> Is that a matter of not running tests locally or a difference between CI and local env?

Yeah, I didn't run all the tests, just `DockerIT`. But CI did its job.

> If the latter, could that be unified?

That's one of the main design choices, we aim to make all tests work the same on CI and locally. Including the performance tests offered by JPT. We do this to empower local development, debugging and freedom from CI limits/queues/costs/permissions, etc.

src/test/kotlin/com/atlassian/performance/tools/infrastructure/DockerIT.kt
```diff
@@ -1,31 +1,53 @@
...
+        return runSoloSsh(ubuntuVersion) { ssh ->
+            DockerImage("hello-world").run(ssh)
         }
     }
 
-    private fun enableNestedDocker(container: GenericContainer<*>) {
-        val dockerDaemonSocket = "/var/run/docker.sock"
-        container.setBinds(listOf(Bind(dockerDaemonSocket, Volume(dockerDaemonSocket))))
+    @Test
+    fun shouldInstallIdempotently() {
```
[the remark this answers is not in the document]
dagguh: Yep, it shouldn't throw. As you the name suggests, it should be idempotent, so running things mutliple times should work.

src/test/kotlin/com/atlassian/performance/tools/infrastructure/DockerIT.kt
```diff
@@ -1,31 +1,53 @@
...
-    fun installWorks() {
-        SshUbuntuContainer(Consumer { enableNestedDocker(it) }).start().use { ssh ->
-            ssh.toSsh().newConnection().use { connection ->
-                //workaround for a bug in Docker download site for bionic
-                val packageFile = "containerd.io_1.2.2-3_amd64.deb"
-                connection.execute("curl -O https://download.docker.com/linux/ubuntu/dists/bionic/pool/edge/amd64/$packageFile", Duration.ofMinutes(3))
-                connection.execute("sudo apt install ./$packageFile", Duration.ofMinutes(3))
+    fun shouldRunOnVariousUbuntuVersions() {
+        AbruptExecutorService(newCachedThreadPool()).use { pool ->
+            listOf("16.04", "18.04", "20.04", "22.04")
```
mgrzaslewicz, line 17: Idea: extract it to constant in api package ,e.g. Ubuntu.supportedVersions and use it where needed
```
/**
 * Runs on Ubuntu {@link com.atlassian.performance.tools.infrastructure.api.supportedUbuntuVersions supported versions}.
 * @param maxConnections MySQL `max_connections` parameter.
 */
class MySqlDatabase(

---
Ubuntu.kt
package com.atlassian.performance.tools.infrastructure.api

val supportedUbuntuVersions = listOf("16.04", "18.04", "20.04", "22.04")

--
DockerIT
  supportedUbuntuVersions
                .map { ubuntuVersion ->
```

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/browser/ChromeIT.kt
```diff
@@ -1,30 +1,24 @@
...
+            val wasInstalledBefore = isChromeInstalled(ssh)
 
-                Assert.assertThat(wasInstalledBefore, Matchers.`is`(false))
-                Assert.assertThat(isInstalledAfter, Matchers.`is`(true))
-                Assert.assertThat(isChromedriverInstalled(connection), Matchers.`is`(true))
-            }
+            Chrome().install(ssh)
 
+            val isInstalledAfter = isChromeInstalled(ssh)
+            Assert.assertThat(wasInstalledBefore, Matchers.`is`(false))
```
mgrzaslewicz, line 19: import static + assertj instead of hamcrest?

dagguh: Agreed, we should standardize on AssertJ. I didn't do in this commit, because the commit was about switching the SSH Ubuntu code. I also didn't include it in the PR itself, because the PR was big enough already. Let's do AssertJ in a separate PR.

mgrzaslewicz:
> That's one of the main design choices, we aim to make all tests work the same on CI and locally. Including the performance tests offered by JPT. We do this to empower local development, debugging and freedom from CI limits/queues/costs/permissions, etc.

We might revisit our approach in other libraries, e.g.
https://github.com/atlassian/jira-software-actions/blob/master/.github/workflows/ci.yml#L19
where we have
```
        jira-version: ["8.13.26", "8.20.13", "9.3.0"]
```
So we depend solely on the CI. On the other hand we run it in CI in parallel, but we might also do that locally

src/test/kotlin/com/atlassian/performance/tools/infrastructure/DockerIT.kt
dagguh: Interesting...
Do you see a use case beyond the tests here?
I'd definitely avoid the implied contract that "`MySqlDatabase` + `AtlassianSplunkForwarder` + `UniversalSplunkForwarder` will always support the same Ubuntu versions".

dagguh:
Correct. Gradle can do that for us locally.

build.gradle.kts
```diff
@@ -66,7 +70,7 @@ dependencies {
     testCompile("com.atlassian.performance.tools:jira-software-actions:[1.0.0,2.0.0)")
     testCompile("org.hamcrest:hamcrest-library:1.3")
     testCompile("org.assertj:assertj-core:3.11.1")
-    testCompile("com.atlassian.performance.tools:ssh-ubuntu:0.2.0")
```
dagguh, line 69: > @mgrzaslewicz: [I'd give it a try with our `ssh-ubuntu` library. It can be easily adjusted to have ubuntu version provided (plus most likely some more modifications)](https://github.com/atlassian/infrastructure/pull/81#issuecomment-1286901107)

> @dagguh [Ok, I'll try on a separate branch.](https://github.com/atlassian/infrastructure/pull/81#issuecomment-1286965166)

I cannot find API to change the version. [The `UbuntuContainer(version = "18.04")` is hardcoded.](https://github.com/atlassian/ssh-ubuntu/blob/release-0.2.6/src/main/kotlin/com/atlassian/performance/tools/sshubuntu/api/SshUbuntuContainer.kt#L20) I'll expose a parameter for that.

mgrzaslewicz: Yup. I was already trying to adjust your solution to use ssh-ubuntu, but got lost in slf4j dependency conflicts

dagguh: Ok, I'll timebox that attempt to 30-45 minutes. I'm applying some time pressure on the alternative, because this JPERF-449 fix blocks a whole chain of other fixes and we want to finish before Jira release.

dagguh: @mgrzaslewicz thanks for the push in the right direction. Here's the [equivalent PR](https://github.com/atlassian/infrastructure/pull/121) based on (updated) ssh-ubuntu. I'm closing this one.

PR #87, opened 2020-11-02
JPERF-619: Reduce max `Ubuntu.install` backoff

dagguh:
There's still 7 retries of `Ubuntu` install.
Exponential backoff with base 5 seconds and exponent of two meant that the 7th attempt took 640 seconds.
Sum of all 7 attempts took 1270 seconds max.
Now the backoff is 5 seconds + jitter between 0 and 10 seconds. Max of 15 seconds per attempt.
Sum of all 7 attempts now takes 105 seconds max.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -48,7 +48,7 @@ dependencies {
 
     implementation("com.atlassian.performance.tools:io:[1.0.0,2.0.0)")
     implementation("com.atlassian.performance.tools:concurrency:[1.0.0,2.0.0)")
-    implementation("com.atlassian.performance.tools:jvm-tasks:[1.0.0,2.0.0)")
+    implementation("com.atlassian.performance.tools:jvm-tasks:[1.2.0, 2.0.0)")
     implementation("org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlinVersion")
     implementation("com.google.guava:guava:23.6-jre")
     implementation("org.apache.httpcomponents:httpclient:4.5.5")
--- a/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/os/Ubuntu.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/os/Ubuntu.kt
@@ -1,12 +1,15 @@
 package com.atlassian.performance.tools.infrastructure.api.os
 
 import com.atlassian.performance.tools.infrastructure.Iostat
-import com.atlassian.performance.tools.jvmtasks.api.ExponentialBackoff
 import com.atlassian.performance.tools.jvmtasks.api.IdempotentAction
+import com.atlassian.performance.tools.jvmtasks.api.JitterBackoff
+import com.atlassian.performance.tools.jvmtasks.api.StaticBackoff
+import com.atlassian.performance.tools.jvmtasks.api.plus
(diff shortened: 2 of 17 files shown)


src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/os/Ubuntu.kt
```diff
@@ -30,9 +33,7 @@ class Ubuntu {
         }
             .retry(
                 maxAttempts = 7, //we need to accommodate cron-based image updates happening in the background
```
dagguh, line 35: @pbruski will 7 attempts spread over 35-105 seconds accommodate that?

dagguh:
CI found a [bug](https://scans.gradle.com/s/65z6fsp74ixds/tests/:testIntegration/com.atlassian.performance.tools.infrastructure.api.os.UbuntuIT/shouldRetry#1).
`JitterBackoff` can result in negative sleeps. Fixing 🛠

dagguh:
Raised https://ecosystem.atlassian.net/browse/JPERF-683

build.gradle.kts
```diff
@@ -48,7 +48,7 @@ dependencies {
 
     implementation("com.atlassian.performance.tools:io:[1.0.0,2.0.0)")
     implementation("com.atlassian.performance.tools:concurrency:[1.0.0,2.0.0)")
-    implementation("com.atlassian.performance.tools:jvm-tasks:[1.0.0,2.0.0)")
+    implementation("com.atlassian.performance.tools:jvm-tasks:[1.1.0, 2.0.0)")
```
mzyromski-atlassian, line 51: Am I right to assume you'll want to release `jvm-tasks` and bump the dependency to `1.1.1` before merging this one?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/os/Ubuntu.kt
pbruski: 35 is cutting it close, download and upgrade can easily take more than that. The original value that caused me problems was 15 seconds IIRC, so 35 is better - but I think a guaranteed min of 60 seconds would be safer. So maybe flip the numbers and make it 10 base + 5 jitter?

build.gradle.kts
dagguh: True! We're blocked (by red CI), which will get unblocked by releasing https://github.com/atlassian/jvm-tasks/pull/6

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/os/Ubuntu.kt
dagguh: Thanks for the context. It's very valuable!

- [ ] ensure minimum 60 seconds total backoff

dagguh: Hmm, to be more precise, 35-105 seconds is the total back-off time alone. If `apt-get update` is particularly slow, its timeout can express our extra patience. But ok, a 60 seconds window of retries (regardless of timeouts) is a nice thing to have anyway.

wyrzyk (approved):

PR #93, opened 2020-12-01
JPERF-664 Use existing package for docker install.

kkulesza2:
Docker.kt was using package that apt-get couldn't find `apt list -a docker` revealed that there is in fact that package, but only with specific version.

Bumping the version solved the problem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,11 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/infrastructure/compare/release-4.17.3...master
 
+### Fixed
++ Bump version of the docker to `5:19.03.13`. Fix [JPERF-664]
+
+[JPERF-664]: https://ecosystem.atlassian.net/browse/JPERF-664
+
 ## [4.17.3] - 2020-11-25
 [4.17.3]: https://github.com/atlassian/infrastructure/compare/release-4.17.2...release-4.17.3
 
--- a/src/main/kotlin/com/atlassian/performance/tools/infrastructure/Docker.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/infrastructure/Docker.kt
@@ -28,7 +28,7 @@ internal class Docker {
         val release = ubuntu.getDistributionCodename(ssh)
         ubuntu.addRepository(ssh, "deb [arch=amd64] https://download.docker.com/linux/ubuntu $release stable", "docker");
 
-        val version = "5:19.03.8~3-0~ubuntu-$release"
+        val version = "5:19.03.13~3-0~ubuntu-$release"
         ubuntu.install(
             ssh = ssh,
             packages = listOf("docker-ce=$version"),


src/main/kotlin/com/atlassian/performance/tools/infrastructure/Docker.kt
```diff
@@ -28,7 +28,7 @@ internal class Docker {
         val release = ubuntu.getDistributionCodename(ssh)
         ubuntu.addRepository(ssh, "deb [arch=amd64] https://download.docker.com/linux/ubuntu $release stable", "docker");
 
-        val version = "5:19.03.8~3-0~ubuntu-$release"
```
dagguh, line 31: If old code was passing tests, then there's no proof the new code fixes a problem

kkulesza2: I reproduced it locally, after changing the version, everything seems to work just fine. And yes there's no proof, that was something that I was trying to tell you during slack conversation. But then I just created snapshot and bumped libraries locally.

mzyromski-atlassian:
Do we know why the problem exposes itself in specific scenarios only? E.g. it doesn't seem to affect our internal Jira regression tests (or at least not constantly) while it makes our cache replication test CI to fail permanently.

kkulesza2:
During installation docker image is taken from repository which is defined in particular infrastructure ( flocal for two infrastructures might be different on AWS and different on our internal regression).

mzyromski-atlassian:
OK, let's give it a try.

mzyromski-atlassian:
It would be great if you could amend the commit and fix the Jira ticket key. Also, a `Fixed` changelog entry would be nice, but we can add it when marking the next release as well.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/Docker.kt
dagguh: Let's say we merge it as is.

Two months pass, another PR comes and updates this value. We'd have to ask you to re-check your local setup.

Or a different PR comes in and completely rewrites the impl (e.g. https://github.com/atlassian/infrastructure/pull/81). he PR author and reviewer don't remember you tested it locally in the past. So they merge it, release it, bump it. Maybe it breaks your local setup.
Undo the bump? But other tests already use the new behavior.
Live on a branch? But it quickly falls behind master, no new features and no bugfixes.
Revert in the lib? But it's already used in many places.

In the long-term, automatic tests are less work.

kkulesza2: Yes, then when the #81 will be merged, conflicts would probably stop the process, so I don't get where the problem is.

I agree that integration tests are desired, but it should be a task that is mainly focused on such things, not single bugfix that will just seal the destiny of one `apt-get install`.

dagguh: This one `apt-get install` is used hundreds of times every month by 5+ teams. It's very important to keep it working.

Conflicts might notice it or not. Or a new PR would not have the conflicts. And even if it is caught then we fall back to option A: asking you to re-test it locally.

CHANGELOG.md
```diff
@@ -28,8 +28,10 @@ Dropping a requirement of a major version of a dependency is a new contract.
 
 ### Fixed
 + Bump MySQL version to 5.7.32. Fix [JPERF-671].
++ Bump version of the docker to `5:19.03.13`. Fix [JPERF-664]
```
mzyromski-atlassian, line 31: This is wrong. We should have marked `5.17.3` release after it was performed. Let's merge https://github.com/atlassian/infrastructure/pull/94 and rebase this branch on `master`.

mzyromski-atlassian (approved):

PR #94, opened 2020-12-03
Mark 4.17.3 release

mzyromski-atlassian:


CHANGELOG.md
```diff
@@ -24,10 +24,13 @@ Adding a requirement of a major version of a dependency is breaking a contract.
...
 ## [Unreleased]
-[Unreleased]: https://github.com/atlassian/infrastructure/compare/release-4.17.2...master
+[Unreleased]: https://github.com/atlassian/infrastructure/compare/release-4.17.3...master
+
+## [4.17.3] - 2020-11-25
+[4.17.3]: https://github.com/atlassian/infrastructure/compare/release-4.17.2...release-4.17.3
 
 ### Fixed
-+ Bump MySQL version to 5.7.32. Fix [JPERF-671].
+- Bump MySQL version to 5.7.32. Address [JPERF-671].
```
dagguh, line 33: True, it wasn't a bug

dagguh (approved):

CHANGELOG.md
mzyromski-atlassian: > True, it wasn't a bug

Btw, doesn't it show incompleteness of CHANGELOG entry types dictionary that we use (Fixed / Added / Changed / Removed / Deprecated)? It's not `Added` as we're indeed not adding any new features and apparently it's also not `Removed` or `Deprecated`. Using a common meaning of `Changed` word, that's the type we're looking at here, but we understand it as changes to the contract, which again is not the case here. But `Fixed` is also not a perfect fit, as we're not fixing a bug.

dagguh: True, it's unintuitive. [keep a changelog](https://keepachangelog.com/en/1.0.0/) does not map to SemVer one-to-one.
I tried to force it to use SemVer semantics. I did that to make "soft mark" MINOR and MAJOR bumps. But we already have to "hard mark" it via release alpha tags. It's just that the maintainer has to know which alpha tag to use after each PR merge.

First principle of keep-a-changelog is:

> Changelogs are for humans, not machines. 

So indeed, we shouldn't use the changelog contents as a helper for SemVer.

Let's decouple the two.

dagguh: PS. https://ecosystem.atlassian.net/browse/JPERF-240 should help with guarding SemVer automatically. It'd cover the source and binary compatibility contracts. Human would only be needed to guard the behavioral and POM contracts.

PR #100, opened 2021-01-29
JPERF-273: Generalize Jira installation API

dagguh:


src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/install/InstalledJira.kt
```diff
@@ -0,0 +1,27 @@
...
+import com.atlassian.performance.tools.infrastructure.api.jvm.JavaDevelopmentKit
+
+/**
+ * Points to an already installed Jira.
+ *
+ * @since 4.18.0
+ */
+class InstalledJira(
+    /**
+     * Contains `./dbconfig.xml` and other server-specific files.
```
mzyromski-atlassian, line 12: It's not clear, at first glance, that the attribute stores a path (as its type is plain String). Maybe we could note that either in the ktdoc or reflect it in the property name?

mzyromski-atlassian (approved):
Great work!

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/install/InstalledJira.kt
dagguh: Good point! I enriched the type.

dagguh:
Resolved conflicts

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/install/ParallelInstallationIT.kt
```diff
@@ -0,0 +1,53 @@
...
+
+        testOnServer { server ->
+            // when
+            val installed = installation.install(server)
+
+            // then
+            val serverXml = installed
+                .installation
+                .resolve("conf/server.xml")
+                .download(Files.createTempFile("downloaded-config", ".xml"))
```
dagguh, line 32: That's what we get when we upgrade a primitive `String` to an object 😃 
Good catch about ambiguity of `installation: String` @mzyromski-atlassian 👍

mzyromski-atlassian (approved):

PR #101, opened 2021-02-12
Generalize Jira start API

dagguh:
Depends on #100

PR #102, opened 2021-02-12
Hook into Jira provisoning

dagguh:
Depends on #101

### DIFF
--- a/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/install/hook/DataCenterHook.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/install/hook/DataCenterHook.kt
@@ -1,0 +1,24 @@
+package com.atlassian.performance.tools.infrastructure.api.jira.install.hook
+
+import com.atlassian.performance.tools.infrastructure.api.jira.SharedHome
+import com.atlassian.performance.tools.infrastructure.api.jira.install.InstalledJira
+import com.atlassian.performance.tools.ssh.api.SshConnection
+
+class DataCenterHook(
+    private val nodeId: String,
+    private val sharedHome: SharedHome
+) : PostInstallHook {
+
+    override fun call(
+        ssh: SshConnection,
+        jira: InstalledJira,
+        hooks: PostInstallHooks
+    ) {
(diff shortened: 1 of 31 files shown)


src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/install/hook/DataCenterHook.kt
```diff
@@ -0,0 +1,24 @@
...
+) : PostInstallHook {
+
+    override fun call(
+        ssh: SshConnection,
+        jira: InstalledJira,
+        hooks: PostInstallHooks
+    ) {
+        val localSharedHome = sharedHome.localSharedHome
+        sharedHome.mount(ssh)
+        val jiraHome = jira.home.path  // TODO what's the difference between localSharedHome and jiraHome? should both be hookable?
```
dagguh, line 19: - [ ] remove the DataCenterHook until we get a DC test

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/report/StaticReport.kt
```diff
@@ -0,0 +1,10 @@
+package com.atlassian.performance.tools.infrastructure.api.jira.report
+
+import com.atlassian.performance.tools.ssh.api.SshConnection
+
+class StaticReport(
+    private val remotePath: String
+) : Report {
+
+    override fun locate(ssh: SshConnection): List<String> = listOf(remotePath)
```
dagguh, line 9: It assumes that the `ssh` will connect to the same server as when it got the `remotePath` during construction.
Might get messy when we get DC hooks support. It'd probably need to return a `List<RemotePath>` and higher layers will have to deal with paths coming from different hosts.

mzyromski-atlassian (approved):
@dagguh, this work is awesome!

dagguh:
Thanks! ❤️ 
I'm working on Data Center version of this. And it seems the abstractions here are insufficient. We need to distinguish a Jira instance from a Jira node. And an instance home from node home (shared home vs local home). I'll put it back into draft. When this is compatible with DC, I will still reduce the scope of the PR to just Server in order to keep it small enough.

dagguh:
Obsoleted by #103

PR #104, opened 2021-04-14
JPERF-732: Download ChromeDriver version that matches installed Chrome version

mzyromski-atlassian:
This is to avoid installing incompatible versions of ChromeDriver and Chrome. For example at the moment, latest ChromeDriver is v90 which requires Chrome v90 which hasn't been released yet.

### DIFF
--- a/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/browser/Chrome.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/browser/Chrome.kt
@@ -1,11 +1,8 @@
 package com.atlassian.performance.tools.infrastructure.api.browser
 
 import com.atlassian.performance.tools.infrastructure.ChromedriverInstaller
-import com.atlassian.performance.tools.infrastructure.ParallelExecutor
 import com.atlassian.performance.tools.infrastructure.api.os.Ubuntu
 import com.atlassian.performance.tools.ssh.api.SshConnection
-import org.apache.http.client.methods.HttpGet
-import org.apache.http.impl.client.HttpClients
 import java.net.URI
 import java.time.Duration.ofMinutes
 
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/browser/Chrome.kt
```diff
@@ -19,19 +16,16 @@ class Chrome : Browser {
...
     }
 
-    private fun getLatestVersion(): String {
-        val httpclient = HttpClients.createDefault()
-        val get = HttpGet("https://chromedriver.storage.googleapis.com/LATEST_RELEASE")
-        val response = httpclient.execute(get)
-        return response.entity.content.bufferedReader().use { it.readLine() }
+    private fun getInstalledMinorVersion(ssh: SshConnection): String? {
+        val versionString = ssh.execute("/usr/bin/google-chrome --version").output
+        return Regex("Google Chrome ([0-9]+\\.[0-9]+\\.[0-9]+)\\.[0-9]+").find(versionString)?.groupValues?.getOrNull(1)
```
dagguh, line 29: Regexes are always scary, but this one is quite readable and the tests are green 👍

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/browser/Chrome.kt
```diff
@@ -19,19 +16,16 @@ class Chrome : Browser {
...
-            { Ubuntu().install(ssh, listOf("google-chrome-stable"), ofMinutes(5)) },
-            {
-                val version = getLatestVersion()
-                ChromedriverInstaller(URI("https://chromedriver.storage.googleapis.com/$version/chromedriver_linux64.zip")).install(ssh)
-            }
-        )
+        Ubuntu().install(ssh, listOf("google-chrome-stable"), ofMinutes(5))
+        val installedMinorVersion = getInstalledMinorVersion(ssh)
+
+        val version = ChromedriverInstaller.getLatestVersion(installedMinorVersion)
```
dagguh, line 22: The driver install is now dependent on the browser, so they can no longer be parallel.
Correctness trumps performance, so that's a good trade. But if we observe provisioning slowdown, this could be the explanation.

dagguh (approved):

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/browser/Chrome.kt
mzyromski-atlassian: Thanks for the comment. I intended to add a similar one, explaining the change, but I got distracted and forgot. :)

PR #107, opened 2021-06-28
JPERF-729: Create a way to configure Jira admin user password during database setup

pczuj:
# This is draft

* I didn't test if SQL is correct
    * I'd like to use code already present in #103 for that, however I'm not yet 100% sure how to do it correctly
* I believe those changes are not compatible with #103 hance I'd like to consult how I can help to reduce the conflicts and possibly contribute to the #103 with the new password overwrite feature in the new form
* I think that this may be a good opportunity to check if changes in #103 are flexible enough for new features like password overwrite in database
* The password overwrite feature could also be implemented with property `-Datlassian.recovery.password=<your-password>`. I decided to use the database setup wrapper without any good reason.


src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
```diff
@@ -0,0 +1,61 @@
...
+        databaseDelegate: Database,
+        sqlClient: SshSqlClient,
+        username: String,
+        encodedPassword: String
+    ) : this(
+        databaseDelegate = databaseDelegate,
+        sqlClient = sqlClient,
+        username = username,
+        encodedPassword = encodedPassword,
+        cwdUserTableName = "jiradb.cwd_user"
```
dagguh, line 35: Builder would contain the defaults and remove all 3 public constructors (leaving just 1 private one), unblocking API evolution.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
```diff
@@ -0,0 +1,61 @@
...
+ * To encode the password use [com.atlassian.crowd.password.encoder.AtlassianSecurityPasswordEncoder](https://docs.atlassian.com/atlassian-crowd/4.2.2/com/atlassian/crowd/password/encoder/AtlassianSecurityPasswordEncoder.html)
+ * from the [com.atlassian.crowd.crowd-password-encoders](https://mvnrepository.com/artifact/com.atlassian.crowd/crowd-password-encoders/4.2.2).
+ */
+class JiraUserPasswordOverridingDatabase(
+    private val databaseDelegate: Database,
+    private val sqlClient: SshSqlClient,
+    private val username: String,
+    private val encodedPassword: String,
+    private val cwdUserTableName: String
+) : Database {
```
dagguh, line 22: That's how `LicenseOverridingMysql` is done, so right now it's ok 👌 

But when #103 comes, the `Database` SPI will be obsolete. [`DockerMysqlServer` is just a `PreInstanceHook`](https://github.com/atlassian/infrastructure/pull/103/files#diff-7d069955c11a4da663e43af1c359b89334fd21dd349e4eb9dfc288ebd25f1c85). But it could expose callbacks to its `SshSqlClient` via its builder.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
```diff
@@ -0,0 +1,61 @@
...
+import com.atlassian.performance.tools.infrastructure.database.SshSqlClient
+import com.atlassian.performance.tools.ssh.api.SshConnection
+import org.apache.logging.log4j.LogManager
+import org.apache.logging.log4j.Logger
+import java.net.URI
+
+/**
+ * Based on https://confluence.atlassian.com/jira/retrieving-the-jira-administrator-192836.html
+ *
+ * To encode the password use [com.atlassian.crowd.password.encoder.AtlassianSecurityPasswordEncoder](https://docs.atlassian.com/atlassian-crowd/4.2.2/com/atlassian/crowd/password/encoder/AtlassianSecurityPasswordEncoder.html)
```
dagguh, line 13: Alternatively, the password could be [transferred via a file](https://github.com/atlassian/infrastructure/blob/release-4.18.0/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/LicenseOverridingMysql.kt#L46-L57), which avoids being logged in SSH logs.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
```diff
@@ -0,0 +1,61 @@
...
+        ssh: SshConnection
+    ) = databaseDelegate.setup(ssh)
+
+    override fun start(
+        jira: URI,
+        ssh: SshConnection
+    ) {
+        databaseDelegate.start(jira, ssh)
+        logger.debug("Changing Jira user $username password to (encoded) $encodedPassword in table $cwdUserTableName")
+        sqlClient.runSql(ssh, "UPDATE $cwdUserTableName SET credential='$encodedPassword' WHERE user_name='$username';")
```
dagguh, line 59: How does Jira know how to decode this? Do they know the salt?

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabaseTest.kt
```diff
@@ -0,0 +1,76 @@
+package com.atlassian.performance.tools.infrastructure.api.database
+
+import com.atlassian.performance.tools.infrastructure.mock.RememberingDatabase
+import com.atlassian.performance.tools.infrastructure.mock.RememberingSshConnection
+import com.atlassian.performance.tools.infrastructure.mock.RememberingSshSqlClient
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+import java.net.URI
+
+class JiraUserPasswordOverridingDatabaseTest {
```
dagguh, line 10: #103 brings integration tests, which cover whole Jira provisioning

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
pczuj: I was trying to understand the rationale behind the SQL file. Is it for the diagnostics purposes to avoid logging?

I'm assuming that the DB stored encrypted passwords are safe to print for the eyes of someone who already has ssh key printed in the log as well.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
pczuj: I don't know, possibly salt is empty, but it's just mu guess

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
dagguh: SSH keys are not logged. JPT logs help with troubleshooting and as such they should be easy to share, without a need for pseudonymisation. So secrets (private keys, passwords, PII) should not be logged. They should be transferred via files if at all.

PS. Raised https://ecosystem.atlassian.net/browse/JPERF-746

dagguh: Logging encrypted secrets is safe only if decryption remains secret. So: https://github.com/atlassian/infrastructure/pull/107#discussion_r659786854

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabaseTest.kt
pczuj: Could I somehow use #103 to write a IT on top of it which I could easily test if the password is correctly replaced (select via sql client as an assert)? Maybe I could use the opportunity to actually contribute to that PR with the feature reimplemented?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
pczuj: TBH I don't like when the logging is part of the class logic. I prefer to create diagnostics mechanisms at the edge of the component usage, that way I'm actually making sure that my implementation is thin enough (similar to how unit tests help model things).

I think I will create a `DiagnosticsSshSqlClient` which will store all of the sql commands in files and it will print those file paths. This will be transparent to the `Database` logic.

pczuj: Or maybe I will just skip the logging

pczuj: > Or maybe I will just skip the logging

Reason is that because the `SshSqlClient` can be overwritten in the `JiraUserPasswordOverridingDatabase`, a wrapper with diagnostics can be written in the test code if the author wants more visibility.

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabaseTest.kt
dagguh: You could target that branch. I push it and rebase it every Friday, but if you merge a PR to it, I will preserve the changes.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
dagguh: The `logger.debug` is up to you. But note that SSH layer is logging all commands anyway.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
pczuj: Builder implemented ✅

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabaseTest.kt
```diff
@@ -52,11 +53,39 @@ class JiraUserPasswordOverridingDatabaseTest {
     }
 
     @Test
-    fun shouldExecuteUpdateOnCwdUserTable() {
+    fun shouldExecuteUpdateOnCwdUserTableWithEncryptedPassword_whenUnknownUserEncryptionMethod() {
```
dagguh, line 56:
```suggestion
    fun shouldUpdateEncryptedPasswordByDefault() {
```

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabaseTest.kt
```diff
@@ -65,14 +94,63 @@ class JiraUserPasswordOverridingDatabaseTest {
...
+        assertThat(sqlClient.getLog())
+            .`as`("sql queries executed")
+            .containsExactly(
+                "select attribute_value from cwd_directory_attribute where attribute_name = 'user_encryption_method';",
+                "UPDATE $cwdUserTableName SET credential='${samplePassword.encrypted}' WHERE user_name='admin';"
+            )
+    }
+
+    @Test
+    fun shouldExecuteUpdateOnCwdUserTableWithEncryptedPassword_whenPlainTextSecurityUserEncryptionMethod() {
```
dagguh, line 121:
```suggestion
    fun shouldUpdatePlaintextPassword() {
```

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabaseTest.kt
```diff
@@ -52,11 +53,39 @@ class JiraUserPasswordOverridingDatabaseTest {
...
+        assertThat(sqlClient.getLog())
+            .`as`("sql queries executed")
+            .containsExactly(
+                "select attribute_value from jiradb.cwd_directory_attribute where attribute_name = 'user_encryption_method';",
+                "UPDATE $cwdUserTableName SET credential='${samplePassword.encrypted}' WHERE user_name='admin';"
+            )
+    }
+
+    @Test
+    fun shouldExecuteUpdateOnCwdUserTableWithEncryptedPassword_whenAtlassianSecurityUserEncryptionMethod() {
```
dagguh, line 84:
```suggestion
    fun shouldUpdateEncryptedPassword() {
```

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/Database.kt
```diff
@@ -8,10 +8,7 @@ import java.net.URI
  */
 interface Database {
 
-    /**
-     * @return Database data location if exists
-     */
-    fun setup(ssh: SshConnection): String
```
dagguh, line 14: This breaks the API.

- [ ] don't break the API unless absolutely necessary (requires major release cascade)

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
```diff
@@ -0,0 +1,97 @@
+package com.atlassian.performance.tools.infrastructure.api.database
+
+import com.atlassian.performance.tools.infrastructure.database.SshMysqlClient
+import com.atlassian.performance.tools.infrastructure.database.SshSqlClient
+import com.atlassian.performance.tools.ssh.api.SshConnection
+import org.apache.logging.log4j.LogManager
+import org.apache.logging.log4j.Logger
+import java.net.URI
+
+data class JiraUserPassword(
```
dagguh, line 10: - [ ] log new API in the changelog

dagguh (changes requested):

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/Database.kt
mgrzaslewicz: Is major release a problem/effort? If not, I would not stop from making small breaking changes to improve the code

wyrzyk: The relese is only one part (the smaller one).  Customers need to adapt to the incompatible changes as well.

wyrzyk: >I would not stop from making small breaking changes to improve the code.

In most cases, you can make changes that are backward compatible.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/Database.kt
```diff
@@ -8,7 +8,10 @@ import java.net.URI
  */
 interface Database {
 
-    fun setup(ssh: SshConnection): DatabaseSetup
+    @Deprecated(message = "Use setupAndGetLocation instead", replaceWith = ReplaceWith("performSetup(ssh)"))
+    fun setup(ssh: SshConnection) = performSetup(ssh).databaseDataLocation
+
+    fun performSetup(ssh: SshConnection): DatabaseSetup
```
dagguh, line 14: Note that this is also breaking API:
[2.4.3. Do not add methods to a released public Java interface
](https://github.com/wyrzyk/java-api-design-checklist/blob/master/CHECKLIST.md#24-interfaces)

You'd have to add a default method. It's problematic in Kotlin, do you'd have to port this class to Java. E.g. we had to do that for [Scenario](https://github.com/atlassian/jira-actions/blob/release-3.16.0/src/main/java/com/atlassian/performance/tools/jiraactions/api/scenario/Scenario.java#L31-L33).

Or we could just forget the `DatabaseSetup` abstraction, because it doesn't bring new semantics/behavior, so it's a speculative abstraction.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/Database.kt
```diff
@@ -8,7 +8,10 @@ import java.net.URI
  */
 interface Database {
 
-    fun setup(ssh: SshConnection): DatabaseSetup
+    @Deprecated(message = "Use setupAndGetLocation instead", replaceWith = ReplaceWith("performSetup(ssh)"))
```
dagguh, line 11: > setupAndGetLocation

But `setupAndGetLocation` doesn't exist.

dagguh, line 11: `ReplaceWith("performSetup(ssh)")` wouldn't actually be a safe replacement, it wouldn't compile.
Instead, we'd need `ReplaceWith("performSetup(ssh).location")`

CHANGELOG.md
```diff
@@ -26,6 +26,14 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/infrastructure/compare/release-4.18.0...master
 
+## [4.19.0] - 2021-11-25
```
dagguh, line 29: This release doesn't exist yet and the date is speculated.
We should leave it under the `Unreleased` section, because that's true.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
dagguh, line 10: Avoid Kotlin data classes in the API. They're impossible to evolve.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
```diff
@@ -0,0 +1,97 @@
...
+        ssh: SshConnection
+    ) = databaseDelegate.performSetup(ssh)
+
+    override fun start(
+        jira: URI,
+        ssh: SshConnection
+    ) {
+        databaseDelegate.start(jira, ssh)
+        if (userPassword == null) {
+            logger.info("Password not provided, skipping dataset update")
```
dagguh, line 42: What is the point of an `JiraUserPasswordOverridingDatabase` instance, which doesn't actually override the password?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/LicenseOverridingMysql.kt
```diff
@@ -89,4 +90,9 @@ internal fun createTempLicenseFile(license: String): File {
         .use { it.write(license) }
     return licenseFile
 
-}
\ No newline at end of file
+}
+
+@Deprecated(message = "Pass licenses as Files to reduce accidental leakage of the license")
```
dagguh, line 95: We shouldn't add API, which is immediately deprecated.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
```diff
@@ -0,0 +1,97 @@
...
+            userPassword = userPassword,
+            cwdUserTableName = cwdUserTableName
+        )
+    }
+
+}
+
+fun Database.withAdminPassword(adminPassword: JiraUserPassword?) = JiraUserPasswordOverridingDatabase.Builder(
+    databaseDelegate = this,
+    username = "admin",
```
dagguh, line 95: This can be a default in the `JiraUserPasswordOverridingDatabase.Builder`, so then we can get rid of this extension.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
```diff
@@ -0,0 +1,97 @@
...
+        jira: URI,
+        ssh: SshConnection
+    ) {
+        databaseDelegate.start(jira, ssh)
+        if (userPassword == null) {
+            logger.info("Password not provided, skipping dataset update")
+        } else {
+            if (shouldUseEncryption(ssh)) {
+                logger.info("Updating credential with encrypted password")
+                sqlClient.runSql(ssh, "UPDATE $cwdUserTableName SET credential='${userPassword.encrypted}' WHERE user_name='$username';")
```
dagguh, line 46: Why is `$cwdUserTableName` a variable, but `cwd_directory_attribute` a constant?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
```diff
@@ -0,0 +1,97 @@
...
+            } else {
+                logger.info("Updating credential with plain text password")
+                sqlClient.runSql(ssh, "UPDATE $cwdUserTableName SET credential='${userPassword.plainText}' WHERE user_name='$username';")
+            }
+            logger.info("Password for user '$username' updated to '${userPassword.plainText}'")
+        }
+    }
+
+    private fun shouldUseEncryption(ssh: SshConnection): Boolean {
+        val dbNamePrefix = if (cwdUserTableName.contains(".")) cwdUserTableName.substringBefore(".") + "." else ""
```
dagguh, line 56: Instead of reverse-engineering the prefix, it's be easier to use and maintain if it was just an explicit parameter.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/LicenseOverridingMysql.kt
```diff
@@ -21,18 +19,21 @@ class LicenseOverridingMysql private constructor(
     private val database: Database,
     private val licenseCollection: LicenseCollection
 ) : Database {
-    private val logger: Logger = LogManager.getLogger(this::class.java)
+    companion object {
```
dagguh, line 22: This exposes a new public API: `LicenseOverridingMysql.Companion`. We can make it private. Or we can let the logger be instance-local.

dagguh (changes requested):

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
mgrzaslewicz: I will change it. Could you give an example?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/LicenseOverridingMysql.kt
mgrzaslewicz: making it private has no effect until kotlin 1.4 :/
https://youtrack.jetbrains.com/issue/KT-11567

Is that why there are no companion objects for loggers in this project? (like private static logger in java initialized once)

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
mgrzaslewicz: This extensions is to easily add this decorator in jpt, I don't think it can be done as easy to read as with extension
```
fun Dataset.withAdminPassword(adminPassword: JiraUserPassword): Dataset {
    return Dataset.Builder(this)
        .database(this.database.withAdminPassword(adminPassword))
        .build()
}
```
and usage
```
    private val adminPassword = getAdminPassword()

    private val dataset = DatasetRegistry.getDataset(
        format = DatasetFormat.EIGHT,
        name = DatasetName.ENTERPRISE_UNIMODAL
    ).withAdminPassword(adminPassword)
```

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
mgrzaslewicz: Makes sense to remove table name parameters if there is no reason to keep it. @pczuj do you remember why you made it as a parameter?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
mgrzaslewicz: It was initially used in DatasetRegistry in jpt to avoid excessive if statements in the code. But now it has been move out of DatasetRegistry so not needed anymore

CHANGELOG.md
```diff
@@ -25,6 +25,11 @@ Dropping a requirement of a major version of a dependency is a new contract.
 
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/infrastructure/compare/release-4.18.0...master
+### Added
+- `JiraUserPasswordOverridingDatabase` to support providing custom admin password during database setup [JPERF-729]
+
+### Deprecated
+- `Database.setup(ssh: SshConnection): String` in favor of `Database.performSetup(ssh: SshConnection): DatabaseSetup`
```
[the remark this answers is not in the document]
pczuj: BTW I'm not yet sure if we shoudl reuse the issue key as it's an internal one. Maybe we want to create a public facing clone.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
pczuj: I think that this kind of "wrapping as a function" API is fine to add. Kotlin stdlib seems to be using similar approach for stream operations on collections.

I agree though that we should move the `username = "admin",` to defaults of the builder and not have it here.

pczuj (commented):
oh... github has this option that makes your comments invisible... cool option. I was wondering why my messages are ignored. Had to click "finish your review"

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
```diff
@@ -0,0 +1,78 @@
...
+
+import com.atlassian.performance.tools.infrastructure.database.SshMysqlClient
+import com.atlassian.performance.tools.infrastructure.database.SshSqlClient
+import com.atlassian.performance.tools.ssh.api.SshConnection
+import org.apache.logging.log4j.LogManager
+import org.apache.logging.log4j.Logger
+import java.net.URI
+
+data class JiraUserPassword(
+    val plainText: String,
```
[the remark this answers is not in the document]
mgrzaslewicz: Because:
- this class is used not only in infrastructure but also in JPT and both plain text + encrypted are needed there
- plain text or encrypted is needed for updating dataset. It depends on crowd configuration (user_encryption_method attribute in cwd_directory_attribute table)

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
```diff
@@ -0,0 +1,78 @@
...
+        } else {
+            sqlClient.runSql(ssh, "UPDATE $cwdUserTableName SET credential='${userPassword.encrypted}' WHERE user_name='$username';")
+            logger.info("Password for user '$username' updated to '${userPassword.plainText}'")
+        }
+    }
+
+    class Builder(
+        private var databaseDelegate: Database,
+        private var username: String,
+        private var userPassword: JiraUserPassword?
```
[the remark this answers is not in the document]
mgrzaslewicz: It's not optional anymore. It made sense when it was used in DatasetRegistry, but it's no longer there so when you use this decorator you definitely want to have a password to be changed

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/LicenseOverridingMysql.kt
```diff
@@ -21,18 +19,19 @@ class LicenseOverridingMysql private constructor(
     private val database: Database,
     private val licenseCollection: LicenseCollection
 ) : Database {
-    private val logger: Logger = LogManager.getLogger(this::class.java)
+    private companion object {
+        private val logger: Logger = LogManager.getLogger(LicenseOverridingMysql::class.java)
```
pczuj, line 23: I don't think we are using static loggers anywhere else (or at least in majority places it is just a class field). Possibly we should follow single approach and I'd suggest to use a field - it probably has _some_ performance overhead on object initialisation, but it also has less boilerplate.

CHANGELOG.md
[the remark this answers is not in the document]
mgrzaslewicz: All previous JPERF links were pointing to ecosystem too. It doesn't mean this one should follow. 
By public which address do you mean?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
```diff
@@ -0,0 +1,88 @@
...
+    private val logger: Logger = LogManager.getLogger(JiraUserPasswordOverridingDatabase::class.java)
+
+    override fun setup(ssh: SshConnection): String = databaseDelegate.setup(ssh)
+
+    override fun start(
+        jira: URI,
+        ssh: SshConnection
+    ) {
+        databaseDelegate.start(jira, ssh)
+        if (shouldUseEncryption(ssh)) {
```
pczuj, line 37: #### Suggestion
(Disclaimer: I understand that the way this class is implemented is a norm that we are following in Atlassian, so I'm not insisting of any change here. It's just a suggestion)

The original version of the `JiraUserPasswordOverridingDatabase` had single responsibility - to update a database entry with given value. It didn't even know what the table name is (ok there was a default, but the idea was that this is not the class responsibility)

Now the class not only has the table name knowledge, but also does some decisions around encryption method.

My suggestion would be to split this into multiple wrappers, where one of them would be the dumb one that just does the update. You can also crate another wrapper which would be a given structure of wrappers initialized at once with single constructor.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
```diff
@@ -0,0 +1,88 @@
...
+            logger.info("Updating credential with encrypted password")
```
pczuj, line 38: Note: The convention in JPT is that INFO log is for the regular user, so that it appears in console output (DEBUG is for power user and it's in file). It's expected that only the key events are logged as INFO.

Subjective: I'm fan of observability via composition. You can have dedicated implementations of wrappers which only responsibility is to do the logging. Here for such wrapper we would need to pass the description of executed query to the next layer, so that the logging wrapper could use that information. Maybe it'd be good to create a new interface for SqlClient which would include such information (as optional?).

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/LicenseOverridingMysql.kt
```diff
@@ -21,18 +19,17 @@ class LicenseOverridingMysql private constructor(
     private val database: Database,
     private val licenseCollection: LicenseCollection
 ) : Database {
-    private val logger: Logger = LogManager.getLogger(this::class.java)
+    private val logger: Logger = LogManager.getLogger(LicenseOverridingMysql::class.java)
```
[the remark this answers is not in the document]
mgrzaslewicz: Forgot to restore it when removed companion object

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
mgrzaslewicz: Maybe I don't understand who the consumers of the library might be.
I thought it's for JPT and provisioning Jira - which means table names are constant.
Where else might this class be used?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/LicenseOverridingMysql.kt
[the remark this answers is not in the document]
pczuj: BTW in the past we had situations where incorrect class was used for logger. This is why we switched to `this::class.java`: https://stash.atlassian.com/projects/JIRASERVER/repos/jira-performance-tests/commits/0abc9352ac35db7ea853c50da5fea48f3bc8f90a#src/test/kotlin/com/atlassian/jira/test/performance/infrastructure/MySqlDatabase.kt

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
pczuj: The logic is fine. What my suggestion is about is the code structure. 

It may be subjective, but I find more granular classes with as limited responsibilities as possible easier to grasp. Such modular approach with the usage of composition can help with the future refactoring, as you may for example be interested in inserting logic between the decision around `shouldUseEncryption` and the actual SQL operation. If we continue to add such logic into a single class we will end up with Jira's "Manager" problem, where all logic is in a single class and it's easy to get lost.

In general it's IMHO good idea to promote [SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle). Frequently new devs who start working on projects are mostly copying patterns existing in the project. Even if the decomposition wouldn't have much practical value directly in our case it would still be worth to do it just for pattern promotion.

CHANGELOG.md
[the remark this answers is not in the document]
pczuj: EAN (ecosystem.atlassian.net) is a public instance. JPERF-729 is on EAN, but it's an internal restricted issue right now. We could open it after releasing the fix, however I'm not sure if we are able to do that.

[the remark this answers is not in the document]
mgrzaslewicz: image
It seems that can be changed easily (well, I didn't click it but I suppose it wouldn't show up if clicking would result in permission failure)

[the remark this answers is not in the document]
pczuj: Yes, technically that can be done. I was more concerned about it from the process perspective.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
pczuj: The reason was to split the responsibility from a class that knows only the structure of table, but doesn't know it's name. It was also to allow overwrites in case the table name changes in the future or is incorrect for some DBs. We are relying on something that is not really an Jira API here.

We could make the fields also configurable for better portability, but then the class will have nothing specific about Jira DB.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordEncryptor.kt
```diff
@@ -0,0 +1,43 @@
+package com.atlassian.performance.tools.infrastructure.api.database.passwordoverride
+
+import com.atlassian.performance.tools.infrastructure.database.SshSqlClient
+import com.atlassian.performance.tools.ssh.api.SshConnection
+import org.apache.logging.log4j.LogManager
+import org.apache.logging.log4j.Logger
+
+interface JiraUserPasswordEncryptor {
+    fun getEncryptedPassword(ssh: SshConnection): String
```
pczuj, line 9: I think we shouldn't have `SshConnection` as an argument for this interface, the ssh is an implementation detail.

If you want to have an interface that is looking for an `SshConnection` I'd suggest to make it a `SshBasedPasswordEncrypterFactory` with `createEncrypter(ssh: SshConnection)` method or something like that.

pczuj: you can also expect lambda `(SshConnection) -> JiraUserPasswordEncryptor` where encrypter's method would be `getEncryptedPassword(password: String)`

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordOverridingDatabase.kt
```diff
@@ -0,0 +1,86 @@
...
+ * from the [com.atlassian.crowd.crowd-password-encoders](https://mvnrepository.com/artifact/com.atlassian.crowd/crowd-password-encoders/4.2.2).
+ */
+
+class JiraUserPasswordOverridingDatabase internal constructor(
+    private val databaseDelegate: Database,
+    private val sqlClient: SshSqlClient,
+    private val username: String,
+    private val jiraDatabaseSchemaName: String,
+    private val userPasswordPlainText: String,
+    private val jiraUserPasswordEncryptor: JiraUserPasswordEncryptor
```
pczuj, line 25: I think this is better now, however I decided to spend some time to show how I'd see the implementation of this mechanism with the current requirements
https://github.com/atlassian/infrastructure/blob/a138b085a9d9acc573a573510d532ada3da13c3a/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraDatabaseReconfigurator.kt

Feel free to ignore it though if you feel like it has no value.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordOverridingDatabase.kt
```diff
@@ -0,0 +1,86 @@
...
+import com.atlassian.performance.tools.infrastructure.database.SshSqlClient
+import com.atlassian.performance.tools.ssh.api.SshConnection
+import org.apache.logging.log4j.LogManager
+import org.apache.logging.log4j.Logger
+import java.net.URI
+
+/**
+ * Based on https://confluence.atlassian.com/jira/retrieving-the-jira-administrator-192836.html
+ *
+ * To encode the password use [com.atlassian.crowd.password.encoder.AtlassianSecurityPasswordEncoder](https://docs.atlassian.com/atlassian-crowd/4.2.2/com/atlassian/crowd/password/encoder/AtlassianSecurityPasswordEncoder.html)
```
pczuj, line 15: This part of doc should probably be adjusted, as the provided password is in plain text and the encryption mechanism is planned to be added.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
dagguh: We can just remove the `data` keyword.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/LicenseOverridingMysql.kt
dagguh: Oh, didn't know that.
The `companion` wasn't used, because we didn't believe that introducing static stuff is necessary. In principle, the logger could be even injectable via builder (with a default).

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
dagguh: We can add a copying constructor for builder, e.g. [like this](https://github.com/atlassian/virtual-users/blob/release-3.13.0/src/main/kotlin/com/atlassian/performance/tools/virtualusers/api/config/VirtualUserBehavior.kt#L152-L154).
PS. The `username` default can still be put into the builder either way.

dagguh: PS. the return type should be `Dataset`

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
dagguh: Then why isn't `cwd_directory_attribute` also parametrized? YAGNI TBH.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/LicenseOverridingMysql.kt
dagguh: Also, premature optimization :wink:

CHANGELOG.md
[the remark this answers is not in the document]
dagguh: We should link the issue even if it's internal. We should/could have a process for revealing the security issues after some time after release.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
dagguh: The update/encrypt responsibility split makes sense. Maybe it would even obsolesce the `JiraUserPassword` type.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
dagguh: True. The "key events" tend to be things that show high level progress or simply long-running processes, like: provisioning Jira, applying load, gathering results.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordEncryptor.kt
dagguh: `SshBasedPasswordEncrypterFactory` -> `SshEncrypterFactory`. They can encrypt any text, not just passwords.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordEncryptor.kt
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    private fun shouldUseEncryption(ssh: SshConnection): Boolean {
+        val sqlResult =
+            sqlClient.runSql(ssh, "select attribute_value from ${jiraDatabaseSchemaName}.cwd_directory_attribute where attribute_name = 'user_encryption_method';").output
+        return when {
+            sqlResult.contains("plaintext") -> false
+            sqlResult.contains("atlassian-security") -> true
+            else -> {
+                logger.warn("Unknown user_encryption_method. Assuming encrypted password should be used")
```
dagguh, line 37: Is it supposed to cover cases when it is missing too?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordEncryptor.kt
```diff
@@ -0,0 +1,43 @@
...
+            userPasswordPlainText
+        }
+    }
+
+    private fun shouldUseEncryption(ssh: SshConnection): Boolean {
+        val sqlResult =
+            sqlClient.runSql(ssh, "select attribute_value from ${jiraDatabaseSchemaName}.cwd_directory_attribute where attribute_name = 'user_encryption_method';").output
+        return when {
+            sqlResult.contains("plaintext") -> false
+            sqlResult.contains("atlassian-security") -> true
```
dagguh, line 35: So `atlassian-security` will work with any `passwordEncryptFunction: (String) -> String`?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordEncryptor.kt
```diff
@@ -0,0 +1,43 @@
...
+import com.atlassian.performance.tools.infrastructure.database.SshSqlClient
+import com.atlassian.performance.tools.ssh.api.SshConnection
+import org.apache.logging.log4j.LogManager
+import org.apache.logging.log4j.Logger
+
+interface JiraUserPasswordEncryptor {
+    fun getEncryptedPassword(ssh: SshConnection): String
+}
+
+internal class DefaultJiraUserPasswordEncryptor(
```
dagguh, line 12: If it's `internal`, then it shouldn't be in API package.
It also means that the only offered `JiraUserPasswordEncryptor` impl cannot be used, so every client would have to implement their own.
PS. what are the use cases for different `JiraUserPasswordEncryptor`s?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordOverridingDatabase.kt
dagguh: I hope the `atlassian-crowd:4.2.2` dependency tree isn't insane :crossed_fingers:
If it is, maybe they have a smaller submodule with the functionality?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/JiraUserPasswordOverridingDatabase.kt
pczuj: I think it's because the table name parameter was made by me with first version of the code was created. Then @mgrzaslewicz took over the PR and his addition didn't follow the same approach. That's where the 2 parts of the code differ.

I don't think YAGNI applies here. As mentioned in my previous comment there are at least 2 reasons for the delegation of table name definition (responsibility split + uncertain contract). Both of those are essentially preparing for easier modification, e.g. it would be easier to adapt the table name from the lib consumer repository rather than relying on a new release.

I'd promote a quite from Matrin Fowler here: https://www.martinfowler.com/bliki/Yagni.html

> Yagni only applies to capabilities built into the software to support a presumptive feature, it does not apply to effort to make the software easier to modify.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordOverridingDatabase.kt
mgrzaslewicz: I thought we decided not to include this dependency in the infrastructure library. I've adjusted documentation

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordEncryptor.kt
mgrzaslewicz: I decided to change that and treat that as an error

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordEncryptor.kt
mgrzaslewicz: no, but I don't want this library to depend on atlassian-security. Or maybe we should include it?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordOverridingDatabase.kt
mgrzaslewicz: I found your implementation a bit too complicated and provided something hopefully easier to understand

.gitignore
```diff
@@ -1,2 +1,3 @@
 .gradle
-build
\ No newline at end of file
+build
+.idea
```
dagguh, line 3: Pro-tip: you can exclude your personal environment files via [global git ignore](https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer)

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordEncryptor.kt
```diff
@@ -0,0 +1,57 @@
...
+) : JiraUserPasswordEncryptor {
+    override fun getEncryptedPassword(plainTextPassword: String) = passwordEncryptFunction(plainTextPassword)
+}
+
+class PlainTextJiraUserPasswordEncryptor : JiraUserPasswordEncryptor {
+    override fun getEncryptedPassword(plainTextPassword: String) = plainTextPassword
+}
+
+
+class DefaultJiraUserPasswordEncryptorProvider(passwordEncryptFunction: (String) -> String) : JiraUserPasswordEncryptorProvider {
```
dagguh, line 49: `DefaultJiraUserPasswordEncryptorProvider` and `DefaultJiraUserPasswordEncryptionTypeService` are very enterprisey. No need to overengineer this, it's just a simple polymorphism with two variants: plain and encrypted. We can do it with one interface and two impls. Especially in the API.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordOverridingDatabase.kt
```diff
@@ -0,0 +1,93 @@
...
+
+}
+
+/**
+ * @param passwordEncryptFunction Based on [retrieving-the-jira-administrator](https://confluence.atlassian.com/jira/retrieving-the-jira-administrator-192836.html)
+ * to encode the password in Jira format use [com.atlassian.crowd.password.encoder.AtlassianSecurityPasswordEncoder](https://docs.atlassian.com/atlassian-crowd/4.2.2/com/atlassian/crowd/password/encoder/AtlassianSecurityPasswordEncoder.html)
+ * from the [com.atlassian.crowd.crowd-password-encoders](https://mvnrepository.com/artifact/com.atlassian.crowd/crowd-password-encoders/4.2.2).
+ *
+ */
+fun Database.withAdminPassword(adminPasswordPlainText: String, passwordEncryptFunction: (String) -> String): Database {
```
dagguh, line 77: This is an alias on the builder, which provides some extra defaults. Let's just put those defaults into the builder.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordEncryptor.kt
```diff
@@ -0,0 +1,57 @@
...
+        return when {
+            sqlResult.contains("plaintext") -> JiraUserPasswordEncryptionType.PLAIN_TEXT
+            sqlResult.contains("atlassian-security") -> JiraUserPasswordEncryptionType.ENCRYPTED
+            else -> throw RuntimeException("Unknown jira user password encryption type")
+        }
+    }
+}
+
+class EncryptedJiraUserPasswordEncryptor(
+    private val passwordEncryptFunction: (String) -> String
```
dagguh, line 39: We shouldn't put Kotlin functino types into the API, because this one is `kotlin.jvm.functions.Function1`, which is clumsy to use from Java. A `java.util.function.Function<String, String>` would do just fine (if we even need anonymous types).

dagguh (changes requested):

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/MySqlDatabase.kt
```diff
@@ -17,7 +17,9 @@ class MySqlDatabase(
     private val source: DatasetPackage,
     private val maxConnections: Int
 ) : Database {
-    private val logger: Logger = LogManager.getLogger(this::class.java)
+    companion object {
+        private val logger: Logger = LogManager.getLogger(MySqlDatabase::class.java)
```
dagguh, line 21: Why bother changing this?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordEncryptor.kt
dagguh: I see. Ok, we can leave the burden of finding the proper function on the consumer for now. We'll see how it goes.

.gitignore
mgrzaslewicz: Good tip. I'd keep it anyway, relying on personal settings doesn't solve the problem for everyone

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordOverridingDatabase.kt
mgrzaslewicz: I'd keep this as it's very convenient in JPT to use it. And it's used in dozens of places
e.g.
```
 private val experimentDataset = PortfolioDataSetProvider
        .forJiraVersion(testConfig.experimentJiraVersion)
        .withAdminPassword(testConfig.adminPasswordPlainText)
```

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/MySqlDatabase.kt
mgrzaslewicz: I made several similar changes (to have static logger init instead of per instance) and then reverted them based on comments in this pull request. This one is a leftover, will revert it

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordOverridingDatabase.kt
dagguh: I see. We can preserve the convenience without duplicating sources of defaults:
```kotlin
 private val experimentDataset = PortfolioDataSetProvider.forJiraVersion(testConfig.experimentJiraVersion)
        .overrideAdmin()  // returns JiraUserPasswordOverridingDatabase.Builder
        .passwordPlainText(testConfig.adminPasswordPlainText)
        // .sqlClient(...)
        // .jiraDatabaseSchemaName()
        .build()
```

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordEncryptor.kt
```diff
@@ -2,56 +2,39 @@ package com.atlassian.performance.tools.infrastructure.api.database.passwordover
...
-
-interface JiraUserPasswordEncryptionTypeService {
-    fun getEncryptionType(ssh: SshConnection, sqlClient: SshSqlClient): JiraUserPasswordEncryptionType
+    fun getEncryptor(ssh: SshConnection, sqlClient: SshSqlClient): JiraUserPasswordEncryptor
 }
 
-class DefaultJiraUserPasswordEncryptionTypeService(
-    private val jiraDatabaseSchemaName: String
-) : JiraUserPasswordEncryptionTypeService {
+class DefaultJiraUserPasswordEncryptorProvider(
```
dagguh, line 15: `Default` doesn't convey the behavior of the implementation (or how it could differ from other possible impls).
This one adapts to `cwd_directory_attribute`, so it could be called `CrowdAdaptiveEncryptorProvider`.

PS. We could drop the "provider" abstraction and make it a `CrowdAdaptiveEncryptor : JiraUserPasswordEncryptor`, delegating to the chosen impl.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordEncryptor.kt
```diff
@@ -2,56 +2,39 @@ package com.atlassian.performance.tools.infrastructure.api.database.passwordover
 
 import com.atlassian.performance.tools.infrastructure.database.SshSqlClient
 import com.atlassian.performance.tools.ssh.api.SshConnection
-
-enum class JiraUserPasswordEncryptionType {
```
dagguh: Nice :smile: bye bye enum :wave:

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordOverridingDatabase.kt
mgrzaslewicz: Still looks more complicated than one line ``.withAdminPassword(testConfig.adminPasswordPlainText)`` and multiple that by all usages - you have more code to maintain

dagguh: It's less code to maintain. See [the diff](https://github.com/pczuj/infrastructure/pull/1/commits/e09bc48692a6679856d32faa393efed40421e914), more red lines than green ones.

dagguh: And it's not more complicated, it gives you one way to pass a given param, instead of multiple ways. Note that the `//` commented lines are not necessary, but they're *possible* now (unlike before).

dagguh:
Here's a PR with some of my suggestions: https://github.com/pczuj/infrastructure/pull/1

dagguh (dismissed):

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserEncryptedPasswordProvider.kt
```diff
@@ -0,0 +1,26 @@
+package com.atlassian.performance.tools.infrastructure.api.database.passwordoverride
+
+import com.atlassian.performance.tools.infrastructure.database.SshSqlClient
+import com.atlassian.performance.tools.ssh.api.SshConnection
+
+interface JiraUserEncryptedPasswordProvider {
+    fun getEncryptedPassword(ssh: SshConnection): String
+}
+
+class CrowdEncryptedPasswordProvider(
```
pczuj, line 10: Would it make sense to move this `CrowdEncryptedPasswordProvider` to separate file?

mgrzaslewicz: done

pczuj: I believe `String#contains` on the SQL query output is not the correct contract for choosing the password encryption. It may be good enough simplification, however I'd suggest to be honest about this hack in the code.

Also in the idea of [SOLID](https://en.wikipedia.org/wiki/SOLID) I'd personally split this class into smaller modules:
 * `SshSqlQueryOutputProvider(sqlClient: SshSqlClient, sqlQueryProvider: () -> String) : (ssh: SshConnection) -> String`
 * `JiraDirectoryAttributeSqlQueryProvider(schema: String, attribute_name: String): () -> String`
 * `JiraDirectoryAttributeSqlQueryOutputParser() : (String) -> List<DirectoryAttribute>`
 * `JiraDirectoryAttributePicker(directoryId: String) : (List<DirectoryAttribute>) -> DirectoryAttribute`
 * `JiraDirectoryAttribute { val directoryId: String, val name: String, val value: String }`
 * `JiraDirectoryAttributeBasedPasswordPicker(passwordMapping: Map<String, String>, defaultHandler: () -> String): (DirectoryAttribute) -> String`

Glueing this together can be done in different ways. The idea is to represent every behavior as class for better modularity. Ideally we could allow every module to be overridable, however the example below doesn't do that.

```
SshSqlQueryExecutor<T>(
    sqlQueryOutputProvider: (ssh: SshConnection) -> String,
    sqlQueryOutputTransformer: (String) -> T
) : (ssh: SshConnection) -> T
```

```
fun <I, T, O> ((I) -> T).map(transformation: (T) -> O) = { input: I -> transformation(this(input)) }
```

```
CrowdEncryptedPasswordProvider(
    sqlClient: SshSqlClient,
    jiraDatabaseSchemaName: String,
    passwordPlainText: String,
    passwordEncryptedWithAtlassianSecurityPasswordEncoder: String
) : (ssh: SshConnection) -> String by SshSqlQueryExecutor(
    sqlQueryOutputProvider = SshSqlQueryOutputProvider(
        sqlClient = sqlClient,
        sqlQueryProvider = JiraDirectoryAttributeSqlQueryProvider(
            schema = jiraDatabaseSchemaName,
            attribute_name = "user_encryption_method"
        )
    )
    sqlQueryOutputTransformer = JiraDirectoryAttributeSqlQueryOutputParser()
        .map(JiraDirectoryAttributePicker("1"))
        .map(
            JiraDirectoryAttributeBasedPasswordPicker(
                passwordMapping = listOf(
                    "plaintext" to passwordPlainText
                    "atlassian-security" to passwordEncryptedWithAtlassianSecurityPasswordEncoder
                ),
                defaultHandler = { throw RuntimeException("Unknown jira user password encryption type") }
            )
        )
)
```
 
Decide yourself it such change could be better or now.

dagguh: We can generate infinite abstractions. The question is: "which ones are useful?". Each abstraction should have a practical (not purely hypothetical) purpose.

mgrzaslewicz: IMHO this is overcomplicating simple job and premature optimization. The example provided is hard to understand, at least for me.
Even extracting JiraUserEncryptedPasswordProvider from JiraUserPasswordOverridingDatabase has some drawback and does not make code easier to understand.
It is already overridable, if someone wants to provide own implementation of password provider one can do it easily.

String result from db contains also some additional lines including header. You can prepare here super precise implementation and do not use String.contains. But what would be the benefit?
Library providing crowd encryption there are 2 enums. If that changes, any implementation will have to be changed.

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordOverridingDatabaseTest.kt
```diff
@@ -0,0 +1,70 @@
...
+class JiraUserPasswordOverridingDatabaseTest {
+
+    private val jira = URI("http://localhost/")
+    private val samplePlainTextPassword = "plain text password"
+    private val expectedEncryptedPassword = "*******"
+
+    private lateinit var database: Database
+    private lateinit var underlyingDatabase: RememberingDatabase
+    private lateinit var sshConnection: RememberingSshConnection
+    private lateinit var sqlClient: MockSshSqlClient
```
pczuj, line 22: Rant / Topic not specific to the PR, but still related to the changes:

I find it somewhat risky to use the industry standard ways of junit test setup. It promotes practices of hiding what is actually happening as part of the code execution. I know that follwoing industry standard reduced WTF/min, however what if the standard itself is incorrect?

An alternative way would be to wrap the test class state into an inner class of the test and return this test state as part of `setup()` factory method. It would cost a single line in each test, however there would be no doubt what the execution order is and there would be no usage of `lateinit var`, so there is less chance that new contributors will repeat it in other places of the code now knowing that we are trying to avoid mutability and inexplicit nullability.

Actionable around it is not required, however I'd like to know you thoughts on that.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordOverridingDatabase.kt
```diff
@@ -0,0 +1,82 @@
...
+
+class JiraUserPasswordOverridingDatabase private constructor(
+    private val databaseDelegate: Database,
+    private val sqlClient: SshSqlClient,
+    private val username: String,
+    private val schema: String,
+    private val userPasswordPlainText: String,
+    private val jiraUserEncryptedPasswordProvider: JiraUserEncryptedPasswordProvider
+) : Database {
+    private val logger: Logger = LogManager.getLogger(this::class.java)
```
pczuj, line 19: Nit: Logging responsibility could be split to a decorator on this class.

mgrzaslewicz: What would be the benefit?

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordOverridingDatabaseTest.kt
dagguh: It's about deduplicating setup code, bit more complex to read, but less to read.

mgrzaslewicz: I've checked 'feelings' around below implementation and the benefit seem to be tiny.
Anyway, I think it might be worth implementing. 
```
class JiraUserPasswordOverridingDatabaseTest {

    private val jira = URI("http://localhost/")
    private val samplePlainTextPassword = "plain text password"
    private val expectedEncryptedPassword = "*******"

    data class TestContext(
        val database: Database,
        val underlyingDatabase: RememberingDatabase,
        val sshConnection: RememberingSshConnection,
        val sqlClient: MockSshSqlClient
    )

    private lateinit var testContext: TestContext

    @Before
    fun setup() {
        val db = RememberingDatabase()
        val sqlClient = MockSshSqlClient()
        testContext = TestContext(
            underlyingDatabase = db,
            sshConnection = RememberingSshConnection(),
            sqlClient = sqlClient,
            database = db
                .overrideAdminPassword(
                    adminPasswordPlainText = samplePlainTextPassword,
                    adminPasswordEncrypted = expectedEncryptedPassword
                )
                .sqlClient(sqlClient)
                .schema("jira")
                .jiraUserEncryptedPasswordProvider(object : JiraUserEncryptedPasswordProvider {
                    override fun getEncryptedPassword(ssh: SshConnection) = expectedEncryptedPassword
                })
                .build()
        )
    }

    @Test
    fun shouldSetupUnderlyingDatabase() {
        // when
        testContext.database.setup(testContext.sshConnection)
        testContext.database.start(jira, testContext.sshConnection)
        // then
        assertThat(testContext.underlyingDatabase.isSetup).`as`("underlying database setup").isTrue()
    }
}
```

mgrzaslewicz: or this way:
```
    @Test
    fun shouldSetupUnderlyingDatabase() {
        with(testContext) {
            // when
            database.setup(sshConnection)
            database.start(jira, sshConnection)
            // then
            assertThat(underlyingDatabase.isSetup).`as`("underlying database setup").isTrue()
        }
    }
```
but both feels weird, maybe because I've never seen nor used such way to implement test just to avoid lateinit var

mgrzaslewicz: So you mean something like that, which requires no knowledge about any magic done by adding @Before annotation
```
class JiraUserPasswordOverridingDatabaseTest {

    private val jira = URI("http://localhost/")
    private val samplePlainTextPassword = "plain text password"
    private val expectedEncryptedPassword = "*******"

    data class TestContext(
        val database: Database,
        val underlyingDatabase: RememberingDatabase,
        val sshConnection: RememberingSshConnection,
        val sqlClient: MockSshSqlClient
    )

    fun setup(): TestContext {
        val db = RememberingDatabase()
        val sqlClient = MockSshSqlClient()
        return TestContext(
            underlyingDatabase = db,
            sshConnection = RememberingSshConnection(),
            sqlClient = sqlClient,
            database = db
                .overrideAdminPassword(
                    adminPasswordPlainText = samplePlainTextPassword,
                    adminPasswordEncrypted = expectedEncryptedPassword
                )
                .sqlClient(sqlClient)
                .schema("jira")
                .jiraUserEncryptedPasswordProvider(object : JiraUserEncryptedPasswordProvider {
                    override fun getEncryptedPassword(ssh: SshConnection) = expectedEncryptedPassword
                })
                .build()
        )
    }

    @Test
    fun shouldSetupUnderlyingDatabase() {
        val testContext = setup()
        with(testContext) {
            // when
            database.setup(sshConnection)
            database.start(jira, sshConnection)
            // then
            assertThat(underlyingDatabase.isSetup).`as`("underlying database setup").isTrue()
        }
    }
}
```

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserEncryptedPasswordProvider.kt
pczuj: > We can generate infinite abstractions. 

I'm not pointing abstractions here. There is limited number of responsibilities and this class has more than one. At least query execution and excetional result condition handling IMHO should be split.

If we are assuming a contract that is not given to us explicitly I'd prefer to have this contract isolated as much as possible. In our case we are assuming that:
 * The query will return only one entry or the first entry returned will be the one that we are interested in - In the example code I added explicit picker for the entry that we are interested in. We could implement it as "take the first one" and document it as a hack. There are also other ways to implement it more explicitly.
 * The query will return header not containing the "plaintext" or "atlassian-security" - In the example code  I added parser for SQL output. This one also may be a hack that checks exactly for `String#contains` and maps it e.g. to an enum class. Such class again would allow us to say in javadoc that it's a hack to reduce cost of the implementation.

> IMHO this is overcomplicating simple job and premature optimization.

Ability to split it to many responsibilities proves that this is not so simple and fact that I'm able to use this way to spot some implicit contracts that the code starts to rely on makes it not premature optimization, but rather quality assurance. 

Still I'm not insisting in chaging it if both of you don't see value in it. So far my understanding of SRP was that we should split as much as possible and that SRP is something that we are following. Possibly one of those is incorrect.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordOverridingDatabase.kt
pczuj: There are multiple benefits:
 * Decoupling logging from a class ensures that your code is observable
     * It's possible to create code that will replace logging with something easier to understand by machine, e.g. event system.
     * It's possible to write unit tests on the same signals as would trigger log lines seen by the user, i.e. you don't need to mock logger to use same signals.
 * You could easily remove the logger code from the execution path by removing the decorator from the initialisation. It could be done dynamically.
 * You could unit test the logger responsibility.
 * It spreads the foundation of the good practice of keeping things SRP.

Again I'm not insisting, especially with this one, because our practice was so far to not decouple logger from other  business logic. Still I wanted to point it out to see your thoughts on it. I personally would prefer us to follow the decoupling practices, because I see how not following it damages the flexibility of our implementations. I don't believe in popular thinking that everything can be rewritten to match the new requirements - from my experience it never happens. However we can write code that is modular and later one be able to initialise it differently to achieve new requirements.

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordOverridingDatabaseTest.kt
pczuj: Yes @mgrzaslewicz exactly that. The junit magic seems to be unnecessary for me.

 Not sure if we should follow this practice in this PR though. Maybe the risk that someone new will come and do it with `@Setup` anyway is the reason to not go for it. I think that having 2 ways in the code may be worse then having only the worse version of the solution. Possibly a separate PR that changes every occurance of `@Setup` to something like this would be correct way.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordOverridingDatabase.kt
mgrzaslewicz: I understand all the benefits, but do we really need them?
Was there a single time in this library we had to test what's being logged or a need to make it configurable?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordOverridingDatabase.kt
```diff
@@ -0,0 +1,82 @@
...
+
+}
+
+/**
+ * @param adminPasswordEncrypted Based on [retrieving-the-jira-administrator](https://confluence.atlassian.com/jira/retrieving-the-jira-administrator-192836.html)
+ * to encode the password in Jira format use [com.atlassian.crowd.password.encoder.AtlassianSecurityPasswordEncoder](https://docs.atlassian.com/atlassian-crowd/4.2.2/com/atlassian/crowd/password/encoder/AtlassianSecurityPasswordEncoder.html)
+ * from the [com.atlassian.crowd.crowd-password-encoders](https://mvnrepository.com/artifact/com.atlassian.crowd/crowd-password-encoders/4.2.2).
+ *
+ */
+fun Database.overrideAdminPassword(adminPasswordPlainText: String, adminPasswordEncrypted: String): JiraUserPasswordOverridingDatabase.Builder {
```
dagguh, line 76: Passing the builder opens up ways to override other params, thanks :+1:

dagguh (approved):

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordOverridingDatabase.kt
dagguh: @pczuj I think these are interesting dev experiments, they deserve to be explored. However, we shouldn't mix them inside a big and overdue feature PR.

PR #112, opened 2022-03-30
PERF-750: remove reference to jira-soke-tests-eu-s3 bucket

gregrow:
The https://s3.eu-central-1.amazonaws.com/jira-soke-tests-eu/ S3 bucket - which must not be used anymore - was only referenced when com.atlassian.performance.tools.infrastructure.api.dataset.HttpDatasetPackage deprecated three-parameter secondary constructor was used - resulting in ObsoleteHttpDatasetPackage instance created. But the download() operation for such-way constructed HttpDatasetPackage would fail anyways due to missing file in the deprecated jira-soke-tests-eu bucket. Therefore, I'm replacing the download() method at ObsoleteHttpDatasetPackage with throw UnsupportedOperationException implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ Adding a requirement of a major version of a dependency is breaking a contract.
 Dropping a requirement of a major version of a dependency is a new contract.
 
 ## [Unreleased]
+### Security
+- Remove `jira-soke-tests-eu` AWS S3 bucket reference
+
+[JPERF-750]: https://ecosystem.atlassian.net/browse/JPERF-750
 
 ## [4.19.0] - 2022-01-17
 [4.19.0]: https://github.com/atlassian/infrastructure/compare/release-4.18.0...release-4.19.0
--- a/src/main/kotlin/com/atlassian/performance/tools/infrastructure/dataset/ObsoleteHttpDatasetPackage.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/infrastructure/dataset/ObsoleteHttpDatasetPackage.kt
@@ -1,12 +1,7 @@
 package com.atlassian.performance.tools.infrastructure.dataset
 
 import com.atlassian.performance.tools.infrastructure.api.dataset.DatasetPackage
-import com.atlassian.performance.tools.infrastructure.api.dataset.FileArchiver
-import com.atlassian.performance.tools.jvmtasks.api.ExponentialBackoff
-import com.atlassian.performance.tools.jvmtasks.api.IdempotentAction
-import com.atlassian.performance.tools.jvmtasks.api.TaskTimer
 import com.atlassian.performance.tools.ssh.api.SshConnection
-import java.net.URI
 import java.time.Duration
 
 @Deprecated("The adapter can be removed with a major release.")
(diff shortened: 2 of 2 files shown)


gregrow:
I was also considering removal of ObsoleteHttpDatasetPackage -- see https://github.com/atlassian/infrastructure/pull/111 or even removal of this 3-arg constructor in com.atlassian.performance.tools.infrastructure.api.dataset.HttpDatasetPackage, but that would require a major release, which I thin we rather avoid now

pczuj (dismissed):

CHANGELOG.md
```diff
@@ -24,6 +24,10 @@ Adding a requirement of a major version of a dependency is breaking a contract.
 Dropping a requirement of a major version of a dependency is a new contract.
 
 ## [Unreleased]
+### Security
+- `jira-soke-tests-eu` AWS S3 bucket reference removed
```
pczuj, line 28: Nit: Changelog entries used to be written in imperative.

pczuj:
I've seen problems with stability of JDK download in JPT before. Probably it would be worth to implement some kind of retry mechanism for it. It shouldn't be hard.

pczuj (approved):

gregrow:
> I've seen problems with stability of JDK download in JPT before. Probably it would be worth to implement some kind of retry mechanism for it. It shouldn't be hard.

Yes, it would help ephemeral outages. This time the outage looks persistent.

gregrow:
> Yeah, Oracle servers were flaky for years. We should rethink whether we should offer unstable capabilities like we do now. That's why internally we have `AtlassianHostedJDK` as a cache in S3. If we expose such a cache in `aws-infrastructure`, then we can move `OracleJDK` there as well, recommending (or hardcoding) the cache and testing with the cache.

@dagguh , right, but I'd suggest checking license compliance, as exposing a cached `OracleJDK` binaries publicly may be a form or _redistribution_

dagguh:
True, but we wouldn't expose a public cache. We'd expose a code to automatically set up an internal cache on your own S3.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/dataset/ObsoleteHttpDatasetPackage.kt
```diff
@@ -16,21 +11,13 @@ internal class ObsoleteHttpDatasetPackage(
...
         ssh: SshConnection
     ): String {
-        val unzipCommand = FileArchiver().pipeUnzip(ssh)
-        TaskTimer.time("download") {
-            IdempotentAction("download dataset") {
-                ssh.execute("wget -qO - ${datasetBucket.resolve(downloadPath)} | $unzipCommand", downloadTimeout)
-            }.retry(5, ExponentialBackoff(Duration.ofSeconds(2)))
-        }
-        return unpackedPath!!
+        throw UnsupportedOperationException("Download of this HttpDatasetPackage (downloadPath='$downloadPath') could not be executed. Create HttpDatasetPackage instance using the two-parameters constructor rather than the deprecated three-parameters one.")
```
mgrzaslewicz, line 17: Would it be better to fail faster in constructor instead of download method? Would save a lot of time for someone running the code

mzyromski-atlassian (approved):

src/main/kotlin/com/atlassian/performance/tools/infrastructure/dataset/ObsoleteHttpDatasetPackage.kt
gregrow: To fail-fast is good, but bear in mind backwards compatibility. As I wrote in the description "the download() operation for such-way constructed HttpDatasetPackage would fail anyways due to missing file in the deprecated jira-soke-tests-eu bucket. Therefore, I'm replacing the download() method at ObsoleteHttpDatasetPackage with throw UnsupportedOperationException implementation."
If someone's code does only construct such HttpDatasetPackage, but never download it, nether the missing file in S3 nor this change should break such code. Well, it's strange to construct HttpDatasetPackage without downloading it, still it's possible.

dagguh: @gregrow I love your empathy towards our lib consumers. Behavioral compatibility is something worth preserving. We even officially promise it in each JPT lib public API definition (see the [CHANGELOG.md compatibility](https://github.com/atlassian/infrastructure/blob/master/CHANGELOG.md) section):
> [behavioral compatibility](http://cr.openjdk.java.net/~darcy/OpenJdkDevGuide/OpenJdkDevelopersGuide.v0.777.html#behavioral_compatibility) with behavioral contracts expressed via Javadoc

We've wanted a reasonable middle ground for what we promise regarding behavior. It's not as free as "we can change anything". It's not as restrictive as "we'll preserve JVM behavior for any possible set of inputs and environments".
Now, we never promised (in JavaDoc) that the `HttpDatasetPackage` constructors does not throw. So technically we're not breaking compatibility. But we can use our empathy (again props to @gregrow) to think whether it should have been promised. In that case, we should make it explicit, so that the consumers feel they have more to rely on and for maintainers to have an easier decision to make.

PS. JavaDoc is not just JavaDoc comments. Class and method names are part of this too. The names promise a certain behavior. And **if** (and that's a big if) we want to promise something more wide-spread like "no constructors are supposed to throw", then we can put it in package/module level JavaDoc or even in the compatibility docs directly.

dagguh: PPS. IMHO we can throw here, because consumers are informed how they can fix the problem through the exception message.

PR #116, opened 2022-06-02
Migrate to artifact publish token

rinconjc:
Artifactory secrets are being deprecated.

Please review and merge this change to use the new artifact publish token.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,9 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
     steps:
     - name: Checkout
       uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


dagguh:
@atlassian/jpt after merging you can release a patch version, even without any changes.

pczuj (approved):

pczuj:
> https://github.com/orgs/atlassian/teams/jpt after merging you can release a patch version, even without any changes.

Should we mark it somehow in CHANGELOG? Would be good for the external consumers to be aware of what the version is for.

pczuj:
I merged counterpart of this in [jira-actions]() module and I'm releasing that. If everything will be fine I will proceed with the rest of the modules (all 7 at once). I will correct the CHANGELOGs later when we will agree on what to put there - I will prepare some draft with my suggestion.

pczuj:
Released 4.19.3 after this change. Everything worked ✅

dagguh:
Sure, the consumers should know that this release exist and that it doesn't contain any changes

PR #119, opened 2022-10-18
JPERF-838: Add Splunkforwarder.jsonifyLog4j method for Log4j1 and Log4j2

filip-nowak-atl:
Jsonify both log4j 1 and log4j 2 config files


CHANGELOG.md
```diff
@@ -26,6 +26,9 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/infrastructure/compare/release-4.19.3...master
 
+### Added
+- Added `jsonifyLog4j(sshConnection: SshConnection, log4jPropertiesPath: String, log4j2ConfigPath: String)` to `SplunkForwarder` to enable jsonifying log4j2 configurations. [JPERF-838]
```
dagguh, line 30:
```suggestion
- Enable jsonifying log4j2 configurations in `SplunkForwarder`. Resolve [JPERF-838].

[JPERF-838] https://ecosystem.atlassian.net/browse/JPERF-838
```

build.gradle.kts
```diff
@@ -33,6 +33,10 @@ configurations.all {
                 "org.jetbrains:annotations" -> useVersion("15.0")
                 "com.google.code.findbugs:jsr305" -> useVersion("3.0.2")
                 "org.apache.commons:commons-compress" -> useVersion("1.9")
+                "net.bytebuddy:byte-buddy" -> useVersion("1.12.14")
+                "org.seleniumhq.selenium:selenium-api" -> useVersion("3.11.0")
+                "org.seleniumhq.selenium:selenium-support" -> useVersion("3.11.0")
+                "org.seleniumhq.selenium:selenium-chrome-driver" -> useVersion("3.11.0")
```
dagguh, line 39: Why?

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/splunk/Log4jJsonifierTest.kt
```diff
@@ -0,0 +1,88 @@
...
+        private const val LOG4J2_CONFIG = "jira-log4j2.xml"
+
+        private const val LOG4J1_REPLACEMENT_CMD =
+            "sed -i -r 's/NewLineIndentingFilteringPatternLayout/layout.JsonLayout/g'"
+        private const val LOG4J2_REPLACEMENT_CMD = "sed -i '/\\/PatternLayout/a\\\n" +
+            "            <AtlassianJsonLayout  filteringApplied=\"false\"\\/>\n" +
+            "/PatternLayout/, /\\/PatternLayout/d'"
+    }
+
+    private var mockConnection: SshConnection = Mockito.mock(SshConnection::class.java)
```
dagguh, line 21: We avoid Mockito and all reflection, auto-magic, hard-to-debug libs. Perhaps you can reuse `RememberingSshConnection`?

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/splunk/Log4jJsonifier.kt
```diff
@@ -0,0 +1,48 @@
+package com.atlassian.performance.tools.infrastructure.api.splunk
```
dagguh, line 1: Let's not leak it to the API:
```suggestion
package com.atlassian.performance.tools.infrastructure.splunk
```

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/splunk/SplunkForwarder.kt
```diff
@@ -5,5 +5,6 @@ import com.atlassian.performance.tools.ssh.api.SshConnection
 interface SplunkForwarder {
     fun run(sshConnection: SshConnection, name: String, logsPath: String)
     fun jsonifyLog4j(sshConnection: SshConnection, log4jPropertiesPath: String)
+    fun jsonifyLog4j(sshConnection: SshConnection, log4jPropertiesPath: String, log4j2ConfigPath: String)
```
dagguh, line 8: Existing `SplunkForwarder` implementations will stop compiling. We maintain source compatibility and this is breaking it.
So we should find a different solution or release a major version.

dagguh: Perhaps we can go for a decorator around `SplunkForwarder`:
```kotlin
class Log4j2SplunkForwarder(
    private val delegate: Log4j2SplunkForwarder
) : SplunkForwarder by delegate {

    override fun jsonifyLog4j(sshConnection: SshConnection, log4jPropertiesPath: String) {
        // ...
    }
}
```

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/splunk/Log4jJsonifierTest.kt
```diff
@@ -0,0 +1,88 @@
+package com.atlassian.performance.tools.infrastructure.api.splunk
+
+import com.atlassian.performance.tools.ssh.api.SshConnection
+import org.assertj.core.api.Assertions.assertThatThrownBy
+import org.junit.Test
+import org.mockito.Mockito
+
+
+internal class Log4jJsonifierTest {
```
dagguh, line 9: TBH this doesn't test much. Everything is mocked so this test suite basically reimplements the `Log4jJsonifier`.

dagguh: Unfortunately infra doesn't have a nice integration test (yet, [incoming](https://github.com/atlassian/infrastructure/pull/103/files#diff-157e10769bd024cfba0f5672d6cf07ba57216d2fec8a8939e3c7ca466643e2beR41)). But I'd rather leave this untested right now.

dagguh (changes requested):

build.gradle.kts
filip-nowak-atl: Good question, gradle detected conflicts and refused to build anything.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/splunk/SplunkForwarder.kt
filip-nowak-atl: Maybe empty default implementation?

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/splunk/Log4jJsonifierTest.kt
filip-nowak-atl: I can remove the test - this way we'll most likely avoid the fun with dependencies.

build.gradle.kts
dagguh: Ah got it, you wanted to add Mockito dependency, but rewriting all locks tried to bump other deps too and conflicts were detected. You can [update the locks selectively](https://docs.gradle.org/current/userguide/dependency_locking.html#selectively_updating_lock_state_entries) to avoid those bumps.

.gitignore
```diff
@@ -1,3 +1,4 @@
 .gradle
 build
-.idea
\ No newline at end of file
+.idea
+.DS_Store
```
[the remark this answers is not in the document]
mgrzaslewicz: I've seen similar comment many times in various pull requests. It means many people don't have such file and it might help to have it in .gitignore per project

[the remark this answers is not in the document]
dagguh: I always imagined all projects in the world having to list all possible OS- and IDE-generated files: Windows, macOS, Notepad++, IntelliJ, VS Code, Visual Studio, Emacs, etc. etc. And it felt wrong.
But I guess you're right. If we follow the path of least resistance, we'll get the most frequent cases out of the way (macOS + IntelliJ).
OTOH there's a bunch of PR feedback that's repeated, like changelog or keeping compatibility. Perhaps we could use [PR templates](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates#pull-request-templates) for that.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/splunk/SplunkForwarder.kt
```diff
@@ -1,9 +1,15 @@
...
 
 interface SplunkForwarder {
     fun run(sshConnection: SshConnection, name: String, logsPath: String)
     fun jsonifyLog4j(sshConnection: SshConnection, log4jPropertiesPath: String)
     fun getRequiredPorts(): List<Int>
-}
\ No newline at end of file
+}
+
+// Extension method to avoid breaking changes
```
dagguh, line 12: [Extension methods](https://kotlinlang.org/docs/java-to-kotlin-interop.html#package-level-functions) in API look funky in Java as a static: `SplunkForwarderKt.jsonifyLog4j1AndLog4j2()`.
You can achieve the same goal by [default methods](https://kotlinlang.org/docs/java-to-kotlin-interop.html#default-methods-in-interfaces), we've got `-Xjvm-default` already enabled.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/splunk/SplunkForwarder.kt
```diff
@@ -1,9 +1,15 @@
...
+fun SplunkForwarder.jsonifyLog4j1AndLog4j2(sshConnection: SshConnection, log4jPropertiesPath: String, log4j2ConfigPath: String) {
```
dagguh, line 13: Have you tried a [decorator](https://github.com/atlassian/infrastructure/pull/119#discussion_r998080974)?

If you're worried you'd have to hunt down all places where `SplunkForwarder` is created, then don't worry. Because if you planned to call this `jsonifyLog4j1AndLog4j2` in place of [the existing call in `StandaloneNodeFormula`](https://github.com/atlassian/aws-infrastructure/blob/release-2.25.8/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/StandaloneNodeFormula.kt#L88), then you might as well use the decorator there. But it gets better, you can apply the decorator right here [in `JiraNoceConfig.Builder.build`](https://github.com/atlassian/infrastructure/blob/release-4.19.3/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/JiraNodeConfig.kt#L146).

This assumes that the decorator can work with the `sshConnection: SshConnection, log4jPropertiesPath: String`, which are already passed. But that's enough already right? It can infer the `log4j2ConfigPath` with something like this:
```kotlin
val log4j2 = log4jPropertiesPath.replace("log4j.properties", "log4j2.properties")
```
With that the decorator should have all the information to handle both versions. At this point, it could be even done directly in existing impls: `AtlassianSplunkForwarder` and `UniversalSplunkForwarder` (via `Log4Jsonifier`), so not even the decorator would be necessary. It'd be an entirely transparent fix with no new API at all.

dagguh (changes requested):

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/splunk/SplunkForwarder.kt
filip-nowak-atl: Not yet, but when I think about it:
The problem with decorator is that it has no possiblity to pass the log4j 2 config filename, the solution to this problem is kind of 'hardcoding' for Jira, where it would be `log4j2.xml` as you propose. I'll try to rewrite this solution to use the decorator pattern though :) OR, just pass the log4j2 config filename via decorators constructor 🤔

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/JiraNodeConfig.kt
```diff
@@ -40,7 +41,7 @@ class JiraNodeConfig private constructor(
         remoteJmx = remoteJmx,
         jvmArgs = jvmArgs,
         collectdConfigs = collectdConfigs,
-        splunkForwarder = splunkForwarder,
+        splunkForwarder = Log4j2SplunkForwarder("log4j2.xml", splunkForwarder),
```
dagguh, line 44: No need to maintain deprecated APIs. If somebody is still using them and suffering from the fix, they can just switch to non-deprecated API.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/splunk/Log4j2SplunkForwarder.kt
```diff
@@ -0,0 +1,16 @@
...
+import com.atlassian.performance.tools.ssh.api.SshConnection
+import java.nio.file.Paths
+
+class Log4j2SplunkForwarder(log4j2ConfigFileName: String, splunkForwarder: SplunkForwarder) :
+    SplunkForwarder by splunkForwarder {
+
+    private val log4jConfigFileName = log4j2ConfigFileName
+
+    override fun jsonifyLog4j(sshConnection: SshConnection, log4jPropertiesPath: String) {
+        val log4j2ConfigPath = Paths.get(log4jPropertiesPath).resolveSibling(log4jConfigFileName).toString()
```
dagguh, line 13: `resolveSibling` nice!

dagguh (approved):

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/splunk/SplunkForwarder.kt
filip-nowak-atl: Fair point, it supports log4j2 xml config only anyway

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jira/JiraNodeConfig.kt
filip-nowak-atl: Removed this change then

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/Sed.kt
```diff
@@ -13,4 +13,28 @@ class Sed {
         val escapedOutput = output.replace("/", "\\/")
         connection.execute("sed -i -r 's/$escapedExpression/$escapedOutput/g' $file")
     }
-}
\ No newline at end of file
+
+    fun safeReplace(
```
dagguh, line 17: Are those needed externally? If not, then we can make them `internal`.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/splunk/Log4jJsonifier.kt
```diff
@@ -0,0 +1,48 @@
...
+
+    fun jsonifyLog4j1AndLog4j2(sshConnection: SshConnection, log4jPropertiesPath: String, log4j2ConfigPath: String) {
+        val log4j1Result = Sed().safeReplace(
+            connection = sshConnection,
+            expression = "NewLineIndentingFilteringPatternLayout",
+            output = "layout.JsonLayout",
+            file = log4jPropertiesPath
+        )
+        if (!log4j1Result.isSuccessful()) {
+            logger.info("Attempt to jsonify $log4jPropertiesPath was unsuccessful.")
```
dagguh, line 30: FYI by default we print `INFO` to the console. Some tests run 10 benchmarks in parallel, each with 2x Jira DC, each with 2+ Jira nodes (and log4j configs). Perhaps all that info is not necessary for the tester? If we switch to `logger.debug`, they'll still be visible in the log files, but they won't be emitted to the console.

dagguh (approved):

src/main/kotlin/com/atlassian/performance/tools/infrastructure/splunk/Log4jJsonifier.kt
filip-nowak-atl: Done

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/Sed.kt
filip-nowak-atl: Done

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/splunk/Log4j2SplunkForwarder.kt
```diff
@@ -0,0 +1,16 @@
+package com.atlassian.performance.tools.infrastructure.api.splunk
+
+import com.atlassian.performance.tools.infrastructure.splunk.Log4jJsonifier
+import com.atlassian.performance.tools.ssh.api.SshConnection
+import java.nio.file.Paths
+
+class Log4j2SplunkForwarder(log4j2ConfigFileName: String, splunkForwarder: SplunkForwarder) :
```
dagguh, line 7: This is new API and it's a good thing. It's a way to offer the fix to anyone, who's not going through the `JiraNodeConfig.Builder`. @atlassian/jpt, this will be a _minor_ release.

dagguh (approved):

pczuj:
FYI I wasn't convinced that this change wasn't breaking module's API so I checked with [japicmp](https://siom79.github.io/japicmp/) and the report confirmed that minor release if correct ✅ 

Can't attach the report, because github rejects HTML attachements + chrome for some reason crashes when taking full page screenshot ¯\_(ツ)_/¯

PR #121, opened 2022-11-07
JPERF-449: Many Ubuntu versions

dagguh:
`Docker` is not an exposed API, but it does power a few public APIs, see the changelog diff.


src/test/kotlin/com/atlassian/performance/tools/infrastructure/DockerIT.kt
```diff
@@ -1,31 +1,53 @@
...
-                DockerImage("hello-world").run(connection)
+    private fun testDockerInstall(version: String) {
+        SshUbuntuContainer.Builder()
+            .enableDocker()
+            .version(version)
+            .build()
+            .start()
+            .use { ubuntu ->
+                ubuntu.toSsh().newConnection().use { connection ->
+                    DockerImage("hello-world").run(connection)
```
pczuj, line 35: I believe that this test relies on the code throwing an exception in case there was some kind of Docker installation/running problem. Do you know how we could enforce such problem to see if the test would fail in case of it? Maybe we could create a test using it? Or maybe there is some other way to ensure that no explicit assert is still testing something. WDYT?

pczuj (approved):

pczuj:
Changes are very clear. Only problem for me was that gradle locks are doing a lot of noise in the PR. Good that you can toggle/roll up those.

mgrzaslewicz (approved):

src/test/kotlin/com/atlassian/performance/tools/infrastructure/DockerIT.kt
dagguh: We can run `docker logs $container` to assert on hello world output

dagguh:
Modern Gradle has more concise lock format. I'll bump it while we're at it.

dagguh:
As agreed earlier, Gradle bump will happen in a separate PR

PR #124, opened 2022-11-14
JPERF-862: Fix MySQL empty password problem

dagguh:


src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/database/MysqlDatabaseIT.kt
```diff
@@ -0,0 +1,35 @@
...
+import com.atlassian.performance.tools.infrastructure.toSsh
+import com.atlassian.performance.tools.sshubuntu.api.SshUbuntuContainer
+import org.junit.Test
+import java.net.URI
+import java.time.Duration
+
+class MysqlDatabaseIT {
+
+    @Test
+    fun shouldStartMysql() {
```
mgrzaslewicz, line 13: How does this test show empty password problem is fixed? And what was the problem?

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/database/MysqlDatabaseIT.kt
```diff
@@ -0,0 +1,35 @@
...
+        val mysql = MySqlDatabase(
+            HttpDatasetPackage(
+                uri = URI("https://s3-eu-west-1.amazonaws.com/")
+                    .resolve("jpt-custom-datasets-storage-a008820-datasetbucket-1sjxdtrv5hdhj/")
+                    .resolve("dataset-f8dba866-9d1b-492e-b76c-f4a78ac3958c/")
```
pczuj, line 18: FYI the dataset comes from https://github.com/atlassian/aws-infrastructure/blob/master/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/DatasetCatalogue.kt#L67-L69

dagguh:
Done in #144

PR #128, opened 2023-01-18
JPERF-917 Update `curl` command used for downloading Oracle JDK

ewefie:
Curl downloads license page instead of JDK archive so later `tar` command fails to unpack this file. This is a known issue, and for a long time the workaround we used worked correctly, but for some reason, it doesn't anymore.
I'm not sure why `wget` works in this case, maybe it's because there's an option to add the flag `--no-cookies`, while for curl there's no such option but I wasn't able to find strong evidence.


pczuj:
> This is a known issue

Were you able to find any internet article about it?

pczuj:
Officially Oracle seems to support curl: https://www.oracle.com/java/technologies/jdk-script-friendly-urls/
Maybe this is something that should be reported to them.

pczuj (approved):

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/OracleJDK.kt
```diff
@@ -38,7 +38,7 @@ class OracleJDK : VersionedJavaDevelopmentKit {
         for (attempt in attempts) {
             try {
                 connection.execute(
-                    cmd = "curl -s -L -O -H 'Cookie: oraclelicense=accept-securebackup-cookie' -k $jdkUrl",
+                    cmd = "wget -q --no-check-certificate --no-cookies --header 'Cookie: oraclelicense=accept-securebackup-cookie;' $jdkUrl",
```
mgrzaslewicz, line 41: `curl -s -L -O -b "oraclelicense=a" -k https://download.oracle.com/otn-pub/java/jdk/8u131-b11/d54c1d3a095b4ff2b6607d096fa80163/jdk-8u131-linux-x64.tar.gz` works fine,
so it can be `curl -s -L -O -b "oraclelicense=a" -k $jdkUrl`

mgrzaslewicz: and the short options (s, L, O, b) can be replaced with their longer, verbose names for readability

mgrzaslewicz (changes requested):
Code can be simpler with curl

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/OracleJDK.kt
pczuj: With resolved options it would be 
```
curl --silent --location --remote-name --cookie "oraclelicense=a" --insecure https://download.oracle.com/otn-pub/java/jdk/8u131-b11/d54c1d3a095b4ff2b6607d096fa80163/jdk-8u131-linux-x64.tar.gz
```

IMHO those longer options are better, however personally I wouldn't be able to guess what some of them do without looking at the manual:
```
-s, --silent        Silent mode (don't output anything)
-L, --location      Follow redirects (H)
 -O, --remote-name   Write output to a file named as the remote file
-b, --cookie STRING/FILE  Read cookies from STRING/FILE (H)
-k, --insecure      Allow connections to SSL sites without certs (H)
```

BTW we probably should skip `-k` (`--insecure`)

ewefie:
> > This is a known issue
> 
> Were you able to find any internet article about it?

I found many topics about this issue on StackOverflow and other programming forums. Just as an example: https://superuser.com/questions/1466580/unable-to-download-oracle-jdk-8-using-wget-command
https://github.com/easybuilders/easybuild-easyconfigs/issues/4575

mgrzaslewicz (approved):

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/OracleJDK.kt
```diff
@@ -38,7 +38,7 @@ class OracleJDK : VersionedJavaDevelopmentKit {
         for (attempt in attempts) {
             try {
                 connection.execute(
-                    cmd = "curl -s -L -O -H 'Cookie: oraclelicense=accept-securebackup-cookie' -k $jdkUrl",
+                    cmd = "curl --silent --location --remote-name --cookie 'oraclelicense=a' --insecure $jdkUrl",
```
pczuj, line 41: As mentioned in other comment we shouldn't use `--insecure`. It ignores SSL certificate and potentially someone could man-in-the-middle attack us by injecting malicious JDK implementation.

pczuj (changes requested):
Please don't use `--insecure`

mgrzaslewicz (approved):

pczuj (approved):

PR #131, opened 2023-02-06
JPERF-971 Fix Ubuntu concurrent usage

mgrzaslewicz:


dagguh (dismissed):

dagguh:
We had green CI: https://github.com/atlassian/infrastructure/actions/runs/4104945432
I force-pushed a bit of diff-cleaning and restored GPG signatures.

dagguh (approved):

dagguh:
Here's [the diff](https://github.com/atlassian/infrastructure/compare/f260d209a8e03111ea0c2013e75c02880054db85..325fb71dd803b13a84c040434acc89652ad90c63) between [last green CI](https://github.com/atlassian/infrastructure/actions/runs/4104945432) and my [rewritten history CI](https://github.com/atlassian/infrastructure/actions/runs/4105343623/jobs/7082065450). (You can get those commit hashes by expanding the `Checkout` step).
There's no functional difference, but last green CI took 18 minutes, while the new one is spinning for 38 minutes and still not done.
@mgrzaslewicz perhaps my [IP+port fix](https://github.com/atlassian/infrastructure/pull/131/commits/a9bcb521f26905fb65fe8740ce6faf2cfce831ea) was not the answer.

Perhaps there's external flakiness? In a different codebase I've been getting:
```
16:02:15,072 WARN  {: } W: Failed to fetch http://security.ubuntu.com/ubuntu/dists/focal-security/InRelease  Connection failed [IP: 185.125.190.36 80]
W: Some index files failed to download. They have been ignored, or old ones used instead.
16:03:10,437 WARN  {self test : DC 2 nodes - 8.20.14 VS 8.20.14: } W: Failed to fetch http://security.ubuntu.com/ubuntu/dists/focal-security/InRelease  Cannot initiate the connection to security.ubuntu.com:80 (2620:2d:4000:1::19). - connect (101: Network is unreachable) Could not connect to security.ubuntu.com:80 (185.125.190.36). - connect (111: Connection refused) Cannot initiate the connection to security.ubuntu.com:80 (2620:2d:4000:1::16). - connect (101: Network is unreachable) Could not connect to security.ubuntu.com:80 (185.125.190.39). - connect (111: Connection refused) Cannot initiate the connection to security.ubuntu.com:80 (2001:67c:1562::18). - connect (101: Network is unreachable) Could not connect to security.ubuntu.com:80 (91.189.91.39). - connect (111: Connection refused) [IP: 91.189.91.39 80]
W: Some index files failed to download. They have been ignored, or old ones used instead.
```

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/os/UbuntuIT.kt
```diff
@@ -81,26 +78,39 @@ class UbuntuIT {
 
     @Test
     fun shouldBeThreadSafe() {
-        val lock = Object()
-        val concurrency = 5
-        val latch = CountDownLatch(concurrency)
+        val completion = ExecutorCompletionService<Unit>(executor)
+        val readyToUseUbuntu = CountDownLatch(1)
```
pczuj, line 82: I'm trying to understand this test. Maybe I'm missing something obvious, but why is it `count = 1` in `CountDownLatch`? Shouldn't it be `5 + 3`, i.e. `8`?

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/os/UbuntuIT.kt
```diff
@@ -81,26 +78,39 @@ class UbuntuIT {
 
     @Test
     fun shouldBeThreadSafe() {
```
pczuj, line 80: TBH I don't trust this test to indicate thread safety. Can we depend on timings of delegated executions inside an uncontrolled system?

This is more some material to think about rather than direct actionable, but maybe we don't need this test at all. Do we want to test if `apt-get` locks work? If not then maybe we could rely on unit tests. An end-to-end test of combined functionality could cover the real integration check.

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/os/UbuntuIT.kt
mgrzaslewicz: It's fine. All threads wait until it's 0. And the count down is run once to make threads start doing ssh commands at the same time

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/os/UbuntuIT.kt
mgrzaslewicz: > Can we depend on timings of delegated executions inside an uncontrolled system?
That was my concern too at the beginning too. But if you want to test timing, you have to start mocking more and more things that will behave like ubuntu. And we want to test this mechanism on the live system
> maybe we don't need this test at all
We do need it. When synchronization is not working, test fails. We also want to cover any ubuntu version for future changes
> Do we want to test if apt-get locks work?
We don't. We want to test that `Ubuntu` uses apt-get properly

PR #132, opened 2023-02-07
Use GitHub token instead of secrets

dagguh:


dagguh:
Actually #127 is the full fix

PR #142, opened 2023-02-21
JPERF-996: Fix fresh dataset compatibility by adding MySQL arguments from Jira documentation

pczuj:
Some of those args are needed when a new MySQL 5.7 is generated using https://hub.docker.com/_/mysql and https://confluence.atlassian.com/adminjiraserver/connecting-jira-applications-to-mysql-5-7-966063305.html

Problem faced without those arguments was Jira starting with error page claiming:
> The database setup is not supporting utf8mb4
> See out documentation for more information on setting up MySQL 5.7
> [Learn more](https://confluence.atlassian.com/jirakb/change-mysql-connection-url-for-jira-server-874744240.html)

We're adding all of the arguments from docs, because there is no reason not to.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,11 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/infrastructure/compare/release-4.22.4...master
 
+### Fixed
+- Fix fresh dataset compatibility by adding MySQL arguments from Jira documentation. Unblock [JPERF-996].
+
+[JPERF-996]: https://ecosystem.atlassian.net/browse/JPERF-996
+
 ## [4.22.4] - 2023-02-15
 [4.22.4]: https://github.com/atlassian/infrastructure/compare/release-4.22.3...release-4.22.4
 
--- a/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/MySqlDatabase.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/MySqlDatabase.kt
@@ -28,6 +28,23 @@ class MySqlDatabase(
     private val ubuntu = Ubuntu()
 
     /**
+     * Arguments based on [jira docs](https://confluence.atlassian.com/adminjiraserver/connecting-jira-applications-to-mysql-5-7-966063305.html).
+     *
+     * We skip setting `--sql-mode` even though the docs says:
+     * "Ensure the sql_mode parameter does not specify NO_AUTO_VALUE_ON_ZERO".
+     * It's unclear what value should be set and the based on [mysql docs](https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html)
+     * the defaults are good.
+     */
+    private val jiraDocsBasedArgs = listOf(
+        "--default-storage-engine=INNODB",
+        "--character-set-server=utf8mb4",
+        "--innodb-default-row-format=DYNAMIC",
+        "--innodb-large-prefix=ON",
+        "--innodb-file-format=Barracuda",
+        "--innodb-log-file-size=2G"
+    ).joinToString(" ")
+
+    /**
      * Uses MySQL defaults.
      */
     constructor(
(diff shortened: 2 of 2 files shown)


src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/MySqlDatabase.kt
```diff
@@ -27,6 +27,23 @@ class MySqlDatabase(
...
 
+    /**
+     * Arguments based on [jira docs](https://confluence.atlassian.com/adminjiraserver/connecting-jira-applications-to-mysql-5-7-966063305.html).
+     *
+     * We skip setting `--sql-mode` even though the docs says:
+     * "Ensure the sql_mode parameter does not specify NO_AUTO_VALUE_ON_ZERO".
+     * It's unclear what value should be set and the based on [mysql docs](https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html)
+     * the defaults are good.
+     */
+    private val jiraDocsBasedArgs = listOf(
```
dagguh, line 38: Good stuff!

dagguh (approved):

pczuj:
Tests done manually:
 * In `infrastructure` running `./gradlew assemble`
 * From `infrastructure` copy `build/lib/infrastructure-4.22.5-SNAPSHOT.jar`
 * Pase under `jira-software-tests/lib` of our internal tests repository (ijpt)
 * In our internal tests repository run `./gradlew testJiraRegressions`
 * Assert that there are 6 results tables presenting no action errors:
     * 2x Standalone
     * 2x DC
     * 2x "Project Issue View In-depth"

✅

dagguh:
WRT tests, you could try harvesting #124

CHANGELOG.md
```diff
@@ -26,6 +26,11 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/infrastructure/compare/release-4.22.4...master
 
+### Fixed
```
pczuj, line 29: Wasn't sure if it's a fix or addition. Any thoughts?

pczuj:
WRT https://github.com/atlassian/infrastructure/pull/124

It'd help with the overall test situation, however the test doesn't check compatibility with Jira and existing datasets, which is currently my biggest concern.

CHANGELOG.md
dagguh: If you would promise this behavior in Javadoc, then the new contract would be`### Added`

dagguh:
At least it automatically checks if the DB starts, unlike the current lack of automatic testing.

CHANGELOG.md
pczuj: I think It's a bit blurry here, because the contract of `MySqlDatabase` is a bit undefined.

I understand it as a bridge between dataset and Jira. As long as dataset is correct any supported by it Jira should work with it.

The question is what is a correct dataset and here is a problem, because we don't have a trace of how a valid dataset is generated (we know how to modify only). Contract of `DatasetPackage` is not defined.

Probably if I'd want to make the change of this PR correctly I'd need to add the data generation mechanism to `infrastructure`. I looked at the code and I don't think I will be able to do that in short time, so I will skip it for now.

For CHANGELOG I'm leaving "Fixed". Because as the contract of `DatasetPackage` is not really defined I will limit my assumptions to thinking that the dataset represents any valid mysql data dir source (inclusing empty dir, as it's valid as well). In such case this `MySqlDatabase` class should do **everything** that it can to make the data dir work with Jira inclusing executing operations from https://confluence.atlassian.com/adminjiraserver/connecting-jira-applications-to-mysql-5-7-966063305.html. This transitively means that any additional support in this class is a **fix**, as we don't say that we don't support something.

pczuj: With this logic we should ~add~ fix support for empty data dir.

dagguh: > the contract of `MySqlDatabase` is a bit undefined

Currently it's defined as:
* a MySQL database
* implementing `Database` SPI
* which can `start` and `setup(jira: URI, ...)`
* and can do so with a `source: DatasetPackage` and `maxConnections: Int`
* and it is `compatible with Ubuntu 16.04, 18.04, 20.04 and 22.04`

dagguh: > Because as the contract of DatasetPackage is not really defined

Correct, the `@param source` in `MySqlDataset` isn't defined. It's doesn't say whether it can be an empty MySQL datadir. We're free to refine the contract now.

PR #151, opened 2023-03-22
Fix flamegraph file extension

dagguh:
Since the bump, it's not SVG anymore.

### DIFF
--- a/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/profiler/AsyncProfiler.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/profiler/AsyncProfiler.kt
@@ -10,7 +10,13 @@ import java.time.Duration.ofSeconds
 /**
  *  Asynchronous profiler. See https://github.com/jvm-profiling-tools/async-profiler#basic-usage
  */
-class AsyncProfiler : Profiler {
+class AsyncProfiler private constructor(
+    private val startParams: List<String>,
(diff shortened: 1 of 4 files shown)


dagguh:
- [ ] add changelog

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/profiler/AsyncProfiler.kt
```diff
@@ -27,24 +33,55 @@ class AsyncProfiler : Profiler {
...
+                throw Exception("The interval $interval seems to big. Usually it's counted in milliseconds or nanoseconds. Try an interval under a second.")
+            }
+        }
+
+        fun extraParams(vararg extraParams: String) = apply {
+            startParams.addAll(extraParams)
+        }
+
+        fun build(): Profiler {
+            return AsyncProfiler(ArrayList(startParams), outputFile)
```
pczuj, line 84: > ArrayList(startParams)

This should make the `Builder` reusable 👍

pczuj (approved):

mgrzaslewicz (approved):

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/profiler/AsyncProfiler.kt
dagguh: Good catch BTW

PR #157, opened 2023-05-10
JPERF-1103 Give more time to Jstat.

ewefie:
Previously 5 builds out of 7 failed. With the fix 5 builds in a row passed, however, there's still another test (MySqlDatabaseIT#shouldStartMysql) that is failing.


dagguh:
We had a 2/7 chance for green `AdoptOpenJdk11IT` before fthe fix. 2/7 = 28%
So if the fix did nothing, and we saw 1 green test, there'd be 28%^1 = 28% chance for accidental green.
With 2 green tests, we'd have 28%^2 = 7,84% chance for accidental green. With 4 green tests, 28%^4 = 0,61%. It would be unbelievable (<1%) that the greenness is by accident. In other words, we're 99% confident the fix is real.

dagguh (approved):

pczuj (approved):

PR #158, opened 2023-05-10
JPERF-1104 Check if Docker is running before pulling MySQL image

ewefie:


dagguh (dismissed):

src/main/kotlin/com/atlassian/performance/tools/infrastructure/Docker.kt
```diff
@@ -69,5 +72,16 @@ internal class Docker {
      */
     private fun startDockerIfNecessary(ssh: SshConnection) {
         ssh.execute("sudo service docker status || sudo service docker start")
+        IdempotentAction("Check if Docker is running") {
```
dagguh, line 75: It's in `src/main`, so we gotta do `CHANGELOG`.

pczuj (approved):
Seems legit. No tests though :(

dagguh:
Pre-fix sample: 12 CIs, 4 failed with the "cannot connect to Docker daemon". @ewefie please add the exception message to the commit message.

@ewefie did the math: 8/12 accidental green = 67%. 67%^12 = 0,81%, so we need 12 green CI to believe they're not accidental.
image

It's a manual statistical test, not a JUnit test. Perhaps we'll figure out a way to automatically handle such flakiness tests. But first we gotta do it manually often enough to gain familiarity and find rough edges.

ewefie:
All 12 builds passed so it looks like the fix works.

dagguh (approved):

PR #161, opened 2023-05-26
Reset captcha

dagguh:
Sometime databases start with a pre-existing failed login count. Then VUs cannot log in due to captcha.
Those failed logins could be purged from the cached database contents, but we can quickly override it in runtime, without having to touch dataset generation process.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ Changing the license is breaking a contract.
 Adding a requirement of a major version of a dependency is breaking a contract.
 Dropping a requirement of a major version of a dependency is a new contract.
 
+### Added
+- Add `ResetCaptcha.Builder`.
+
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/infrastructure/compare/release-4.24.2...master
 
--- a/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/captcha/ResetCaptcha.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/captcha/ResetCaptcha.kt
@@ -1,0 +1,49 @@
+package com.atlassian.performance.tools.infrastructure.api.database.captcha
+
+import com.atlassian.performance.tools.infrastructure.api.database.Database
+import com.atlassian.performance.tools.infrastructure.database.SshMysqlClient
+import com.atlassian.performance.tools.infrastructure.database.SshSqlClient
+import com.atlassian.performance.tools.ssh.api.SshConnection
+import java.net.URI
+
+
+/**
+ * [Reset a user's login count from the database in Jira server](https://confluence.atlassian.com/jirakb/reset-a-user-s-login-count-from-the-database-in-jira-server-329352344.html)
+ */
+class ResetCaptcha private constructor(
+    private val database: Database,
+    private val username: String,
+    private val sqlClient: SshSqlClient,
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -23,6 +23,9 @@ Changing the license is breaking a contract.
 Adding a requirement of a major version of a dependency is breaking a contract.
 Dropping a requirement of a major version of a dependency is a new contract.
 
+### Added
```
dagguh, line 26: I would add it in iJPT, but `SshSqlClient` is `internal`.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/captcha/ResetCaptcha.kt
```diff
@@ -0,0 +1,49 @@
...
+
+        fun database(database: Database) = apply { this.database = database }
+        fun username(username: String) = apply { this.username = username }
+        fun schema(schema: String) = apply { this.schema = schema }
+
+        fun build() = ResetCaptcha(database, username, sqlClient, schema)
+    }
+}
+
+fun Database.resetCaptcha(): Database = ResetCaptcha.Builder(this).build()
```
dagguh, line 49: It's different from `Database.overrideAdminPassword` extension function, it doesn't return a builder. Extension functions are syntax sugar, so let's actually save boilerplate with it, e.g.:
```kotlin
dataset
    .database
    .resetCaptcha()
    .overrideLicense(...)
```
Instead of:
```kotlin
dataset
    .database
    .resetCaptcha()
    .build()
    .overrideLicense(...)
```
If a more complex use case needs the access to the builder, they can just use the `ResetCaptcha.Builder` directly.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/captcha/ResetCaptcha.kt
```diff
@@ -0,0 +1,49 @@
...
+ */
+class ResetCaptcha private constructor(
+    private val database: Database,
+    private val username: String,
+    private val sqlClient: SshSqlClient,
+    private val schema: String
+) : Database {
+    override fun setup(ssh: SshConnection) = database.setup(ssh)
+
+    override fun start(jira: URI, ssh: SshConnection) {
```
dagguh, line 21: Yup, no tests. It might seem that similar API was tested. But the existing `JiraUserPasswordOverridingDatabaseTest` doesn't actually test anything. It mocks everything and verifies mock interactions. You can break the SUT without tests noticing, e.g. you can move ` databaseDelegate.start(jira, ssh)` **after** `sqlClient.runSql` and the test will pass, but in reality the client would fail, because it's not installed. I could mock everything and assert on trivial logic (2 strings), but it wouldn't give me any extra confidence.

ewefie (approved):

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/captcha/ResetCaptcha.kt
pczuj: I think this aspect is well described in https://www.youtube.com/watch?v=QFCHSEHgqFE, mostly in the "User Contract Testing" section. We may want to watch it on one of our Friday coffee meetings. It's on our list.

PR #174, opened 2023-12-05
Bump async-profiler to v2.10

dagguh:


mgrzaslewicz (approved):

mgrzaslewicz (approved):

dagguh:
Actually, we don't need this bump. We only need to add `interval` when using `wallClockMode`.
Source: https://github.com/async-profiler/async-profiler/pull/748, https://youtrack.jetbrains.com/issue/IDEA-320011

dagguh:
The bump doesn't fix it, but it doesn't break it either, we don't have to revert :+1:

PR #180, opened 2024-06-03
DCPERF-430 Add VersionedOracleJdk

mgrzaslewicz:

PR #183, opened 2024-06-11
Add TarGzJdk to install any jdk

mgrzaslewicz:


src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/TarGzJdk.kt
```diff
@@ -0,0 +1,70 @@
...
+        connection.execute("echo '${use()}' >> ~/.profile")
+    }
+
+    override fun use(): String = "export PATH=$javaHome/bin:\$PATH; export JAVA_HOME=$javaHome"
+
+    override fun command(options: String) = "$javaHome/bin/java $options"
+
+    override val jstatMonitoring = Jstat("$javaHome/bin/")
+
+    class Builder(
```
dagguh, line 43: Right now it's not a builder, but a copy of the constructor.

- [ ] no-arg constructor + defaults

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/TarGzJdk.kt
```diff
@@ -0,0 +1,70 @@
...
+        val archive = "jdk-$version.tar.gz"
+        IdempotentAction("download JDK $version from $downloadUrl") {
+            connection.execute(
+                cmd = "curl --location $downloadUrl > $archive",
+                timeout = ofSeconds(120)
+            )
+        }.retry(3, ExponentialBackoff(ofSeconds(10)))
+        connection.execute("mkdir $javaHome")
+        // unzip to javaHome regardless of the main directory name in the archive
+        val tempDir = connection.execute("mktemp -d").output.lines().first()
```
dagguh, line 29: `mktemp` TIL, nice

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/TarGzJdk.kt
```diff
@@ -0,0 +1,70 @@
...
+            patchVersion: String,
+            downloadUrl: String
+        ) = apply {
+            this.majorVersion = majorVersion
+            this.minorVersion = minorVersion
+            this.patchVersion = patchVersion
+            this.downloadUrl = downloadUrl
+        }
+
+        fun build(): TarGzJdk = TarGzJdk(
```
dagguh, line 62:
```suggestion
        fun build(): TarGzJdk = VersionedJavaDevelopmentKit(
```

src/test/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/TarGzJdkIT.kt
```diff
@@ -0,0 +1,28 @@
+package com.atlassian.performance.tools.infrastructure.api.jvm
+
+import org.junit.Test
+
+class TarGzJdkIT {
```
dagguh, line 5: Nice, thx

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/TarGzJdk.kt
```diff
@@ -0,0 +1,70 @@
...
+    private val version = "$majorVersion.$minorVersion.$patchVersion"
+    private val javaHome = "~/jdk-$version"
+
+    override fun getMajorVersion() = majorVersion.toInt()
+
+    override fun install(connection: SshConnection) {
+        val archive = "jdk-$version.tar.gz"
+        IdempotentAction("download JDK $version from $downloadUrl") {
+            connection.execute(
+                cmd = "curl --location $downloadUrl > $archive",
```
dagguh, line 23: Long opts, thx

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/TarGzJdk.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    override fun use(): String = "export PATH=$javaHome/bin:\$PATH; export JAVA_HOME=$javaHome"
+
+    override fun command(options: String) = "$javaHome/bin/java $options"
+
+    override val jstatMonitoring = Jstat("$javaHome/bin/")
+
+    class Builder(
+        private var majorVersion: String,
+        private var minorVersion: String,
```
pczuj, line 45: Do we need `minorVersion` and `patchVersion`? Maybe we could skip those and make the API simpler.

CHANGELOG.md
```diff
@@ -26,6 +26,9 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/infrastructure/compare/release-4.29.0...master
 
+### Added
+- Add `TarGzJdk` to support any JDK packaged in a tar.gz archive. Help fix [DCPERF-432]
```
dagguh, line 30: So apparently #180 wasn't enough. Something wrong with Oracle JDK?

dagguh (changes requested):

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/TarGzJdk.kt
mgrzaslewicz: I was thinking about it.
It's good to see in the logs full version of JDK path for debugging purposes in lib consumers, not just only '~/jdk-17'.
We could
- pass just `version`, but then parse it to get major version
- pass `majorVersion` only - but then we don't know the descriptive folder name for JDK location
- don't assume javaHome, instead use main dir name. It's the opposite of current way which is assume javaHome then adjust archive content

CHANGELOG.md
mgrzaslewicz: Jira is not starting with Oracle JDK 17. However what we know it's starting with open JDK. I was developing `TarGzJdk` internally but it was lacking a devloop so moved it here. I still don't know the answer if it will help - I'm waiting the e2e test to finish

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/TarGzJdk.kt
pczuj: Passing only `majorVersion` seems fine as it's needed in `VersionedJavaDevelopmentKit` API. For observability/diagnostics can't we print the URL? We should end up with full knowledge in the log anyway.

3rd option seems good.

CHANGELOG.md
dagguh: I see. The famous hooks PR would finally give infrastructure Jira starting capability, a more useful develoop.

Anyway, is it ok for Jira to fail to start with Oracle JDK?

mgrzaslewicz: I'm closing the PR, it doesn't help. I don't want to add it in case we need it - unnecessary complexity. It's left in the history if we need it back

PR #186, opened 2024-06-19
Dcperf 432 install fonts

mgrzaslewicz:


src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/JdkFonts.kt
```diff
@@ -0,0 +1,14 @@
+package com.atlassian.performance.tools.infrastructure.api.jvm
+
+import com.atlassian.performance.tools.infrastructure.api.os.Ubuntu
+import com.atlassian.performance.tools.ssh.api.SshConnection
+
+internal class JdkFonts {
+
+    fun install(ssh: SshConnection) {
+        val packages = listOf(
+            "fontconfig" // https://jira.atlassian.com/browse/CONFSRVDEV-8954
```
dagguh, line 10: https://confluence.atlassian.com/jirakb/jira-server-7-13-or-later-fails-with-fontconfiguration-error-when-installing-on-linux-operating-systems-964956221.html
We can run `install_fontconfig.sh` when installing Jira (if it exists). But doing it "manually" is also ok as per the KB above.

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/AdoptOpenJDK.kt
```diff
@@ -21,6 +21,7 @@ class AdoptOpenJDK : VersionedJavaDevelopmentKit {
         download(connection)
         connection.execute("tar -xzf $jdkArchive")
         connection.execute("echo '${use()}' >> ~/.profile")
+        JdkFonts().install(connection)
```
dagguh, line 24: Changelog please
- [ ] changelog

dagguh (changes requested):

src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/jvm/JdkFonts.kt
dagguh: We can link the KB here tho

dagguh (approved):

